PR #37, opened 2021-04-27
feat: building google analytics plugin

AceTheCreator:
**Description**

- ...
- ...
- ...

**Related issue(s)**


AceTheCreator:
@fmvilas Please kindly check out this draft PR and help me out 🙏

src/lib/client-events.js
```diff
@@ -0,0 +1,10 @@
+const EventEmitter = require("events");
+const clientEvents = module.exports;
+
+const eventEmitter = new EventEmitter();
+
+clientEvents.on = eventEmitter.on.bind(eventEmitter);
+clientEvents.once = eventEmitter.once.bind(eventEmitter);
+clientEvents.emit = (eventName, eventPayload, ...args) => {
+  eventEmitter.emit(eventName, eventPayload, ...args);
+};
```
fmvilas, line 10: I wonder if we can just do this:

```suggestion
module.exports = eventEmitter;
```

yarn-error.log
```diff
@@ -0,0 +1,112 @@
+Arguments:
```
fmvilas, line 1: I think you can get rid of this file 😄 Also, add it to .gitignore so it doesn't happen again.

plugins/google-analytics/src/lib/google-analytics.js
```diff
@@ -0,0 +1,8 @@
+const ReactGA = require('react-ga');
+
+const googleAnalytics = module.exports;
+
+googleAnalytics.init = (window) => {
+    ReactGA.initialize('UA-000000-01');
```
fmvilas, line 6: I wonder how can we make this id configurable without exposing server-side secrets into the client code 🤔

fmvilas (changes requested):
Overall looks great! I think you got the idea 🙌 I left some comments on how to improve it but this is mostly done.

PS: Sorry for the delay. Next time please ping me again so I don't forget 😅

plugins/google-analytics/src/lib/google-analytics.js
AceTheCreator: What if the id can be passed as an extra arg when emitting the event on page render. @fmvilas what do you think?

magicmatatjahu:
@AceTheCreator I copy your comment from issue :)

> > ### Description
> > We used to have Google Analytics hardcoded (see #13). However, some people may not want to use it so it's better if we make it a plugin instead.
> > 
> > ### Implementation suggestion
> > 1. Go to `src/pages/_app.js` file and make it trigger a new kind of event on the `render` function. For instance, something like a `page:render` event. Make sure this event is only triggered on the client-side render.
> > 2. Make the plugin subscribe to the event and trigger an `initialize` and `pageview` call when a new event comes in.
> > 
> > ### Hints
> > The code for triggering/subscribing to events is on `src/lib/events.js`. However, this library was created for the backend and it imports the `config` lib, which may contain secrets. If we use the `events` lib on the frontend, these secrets may be exposed on the browser. Watch out!
>
> @fmvilas when I subscribe to the event from a different file and I trigger it in the app.js render it doesn't work. For instance, when I call event names from the app.js to check the list of subscribed events, it returns an empty array. Except I call both the subscribe and trigger event within the page render function in the app.js
@fmvilas please check out my updated draft PR

The problem is that you don't have any triggers in the client side due to fact that writing code in `render` function under `typeof window !== undefined` you execute code only on client side, it means that binded events to registry are missed during rendering page on the server. NextJS works in this way that prepares data for component (`getInitialProps` function), then these data is transformed to JSON and attached to page. NextJS render page on server side, using of course `getInitialProps` function and then if user go to page React on client side hydrate page using these attached JSON data. Your case is hard, because you try emit event from registry binded on server side and then this data, on client side, is lost. So we must think how to "attach" events from external packages to client side 😢

AceTheCreator:
> The problem is that you don't have any triggers in the client side due to fact that writing code in `render` function under `typeof window !== undefined` you execute code only on client side, it means that binded events to registry are missed during rendering page on the server. NextJS works in this way that prepares data for component (`getInitialProps` function), then these data is transformed to JSON and attached to page. NextJS render page on server side, using of course `getInitialProps` function and then if user go to page React on client side hydrate page using these attached JSON data. Your case is hard, because you try emit event from registry binded on server side and then this data, on client side, is lost. So we must think how to "attach" events from external packages to client side and also how to `import` external plugins for client side, because at the moment it only works for server side (code for that is in `plugins.js` file).

@magicmatatjahu what do you think will be the best approach in tackling this problem?

src/pages/_app.js
```diff
@@ -1,40 +1,41 @@
...
+  if (typeof window !== 'undefined') {
+    clientEvents.on('page:render', () => googleAnalytics.init(window));
+    clientEvents.emit('page:render');
+  }
+  return (
+    <AppContext.Provider value={context || {}}>
+      <Component {...pageProps} />
+    </AppContext.Provider>
+  );
+}
```
magicmatatjahu, line 17: @AceTheCreator I will answer here for you question. 

As I wrote, the code:

```ts
    clientEvents.on('page:render', () => googleAnalytics.init(window));
    clientEvents.emit('page:render');
```

will be only executed on the client side and you will lost all events binded to `clientEvents` registry. As you wrote in previous comments, `googleAnalytics.init` only works when you bind this plugin as `clientEvents.on('page:render', () => googleAnalytics.init(window));` in this file. So I see steps to do:

- load events from plugins to `clientEvents` registry. We must remember that they should be loaded on client side, so loading in this file is probably the best place, but:
  - to load events we use the file system. It doesn't exist on client side, so we must handle this problem on different way, I mean load events from external packages/plugins but without filesystem. We can use https://nextjs.org/docs/advanced-features/dynamic-import#with-no-ssr to load dynamic modules on client side :) However there is an another problem, because we must know about names of external packages/plugins. So this is the hardest part. How to get the packages name, and then load events from them? If you don't understand, please see example:
  
    If we will use the dynamic import from nextJS, we will have something like:
    
    ```js
    const DynamicPlugin = dynamic(
      () => import('@asyncapistudio/plugin-google-analytics'), // I didn't test it, so I don't know if this works, but should
      { ssr: false }
    )
    
    clientEvents.on('page:render', () => DynamicPlugin.init(window));
    ...
    ```
    
    So we have hardcoded name of the plugin in source file. How to get external plugins from `config.plugins` array? To be honest I don't know, because then we must to do:
    
    ```js
    const pluginsName = ... // here we must load the names of plugins from `config.plugins` array, but we cannot use filesystem, because data from it will be lost in client side, maybe some logic in `getInitialProps` will help
    
    for (const pluginName of pluginsName) {
      const DynamicPlugin = dynamic(
        () => import(pluginName), // I didn't test it, so I don't know if this works, but should
        { ssr: false }
      );
      clientEvents.on('page:render', () => ... bind page:render event from plugin);
    }
    ```
    
- Execute `page:render` event.

Let me know if you understand the problem. I really don't know how to solve it the easiest way, because we're trying to bind the filesystem to the client side which is hard to do, I can say it is impossible. I don't see even very complicated solution at the moment for this problem 😢 

I will try to spend some time on this problem, but I don't give guarantee that I will be able to solve something.

magicmatatjahu: @AceTheCreator Hi! No worries, we know that you also study :) It's an open source so we also know that people works in the free time :) Please also don't waste your time on this issue, because it's really hard to make it in the good way. You should focus on your GSOC project :)

magicmatatjahu:
@AceTheCreator I opened issue https://github.com/asyncapi/studio/issues/80 Probably we will remove the NextJS and go with pure React app, so I will close PR, or do you have a different opinion on the subject?

AceTheCreator:
> @AceTheCreator I opened issue #80 Probably we will remove the NextJS and go with pure React app, so I will close PR, or do you have a different opinion on the subject?

@magicmatatjahu It's fine. So will any suggested plugins be built separately?

magicmatatjahu:
@AceTheCreator It depends. From what he remembers, the chatbot needs a token for this AI service (I don't remember name, sorry). If there is a library that can communicate in the browser, then the chatbot could be implemented just for the browser. As for the server API, you could use a websocket that reads the answers and asks the next question based on it :)

magicmatatjahu:
@AceTheCreator The Studio will be SPA application, so there will be plugins - you can read proposal [here](https://github.com/asyncapi/asyncapi-react/issues/433). If you mean plugins for (server) API then we can also reuse my proposal - you can add comment/ask about plugins in the https://github.com/asyncapi/studio/issues/86 issue :)

Issue #80, opened 2021-09-15
Proposal to remove NextJS


magicmatatjahu:
The current studio is written using the [NextJS framework](https://nextjs.org/). The application consists of several parts (because it's not the plain NextJS app):

- server which is running by ExpressJS. Its role is to communicate with the database, save AsyncAPI documents, generate HTML Template as well as Markdown Template etc.
- "Static" UI - the part that communicates with the server and is the UI for it, the part where you manage your user account, save your documents etc, create organizations etc. - Something like https://swagger.io/tools/swaggerhub/ but for AsyncAPI. Of course this part is not that extensive, it has only a few functionalities.
- Dynamic UI - part with editor and where HTML template is rendered. 

The most important part for us is the third part, which is related to the editor and displaying the documentation in various way.

The first part currently exists, but for some reason we can't and don't even want to manage and store data about users, their organizations and the documents themselves. The only thing we would need from the server would be a few endpoints to generate the templates and save it to the user's disk. Just that.

Due to the requirements, the 1st and 2nd part "completely" falls out. That leaves only 3. Personally, I don't see the point of using NextJS where we wouldn't use any of its advantages, i.e. SSR (Server Side Rendering), or SSG (Static Site Generation). Studio should be a frontend application that performs all operations on the client side. Most of our libraries that we currently use (and will use, like diff etc.) in the client-side studio are isomorphic (you can use them client-side and server-side), so that's another reason to remove NextJS. Of course I still see the point of having a server, but for things like using our [generator](https://github.com/asyncapi/generator) that we can't use on the client side. But under this case, a simple SPA application, which communicates to the simple server setup next to it, is enough for us, without any SSR.

I know that the current studio also has the ability to define plugins, but... they only work properly on the server side. So they only work for the first two parts of the Studio, the Server and the Static UI. You can't change or add new things to the editor itself, because initialization of these plugins is at the Node level, not at the browser level. You can read more about this problem here -> https://github.com/asyncapi/studio/pull/37#issuecomment-843366186

Treating the studio itself like a SPA will additionally allow us to publish this as a simple react component -> https://github.com/asyncapi/asyncapi-react/issues/146

Here I have to sorry all ofyou, because I made a mistake and didn't ask anyone in the community if removing NextJS makes sense. I saw so many problems that I thought I was doing the right thing, but AsyncAPI is a community-driven organization, so I apologize. 

What do you think about this? Maybe I'm completely wrong and it makes sense to leave the NextJS.

[...]

PR #88, opened 2021-09-24
feat: implement studio from scratch

magicmatatjahu:
**Description**

- ...

**Related issue(s)**
Proposal outcome https://github.com/asyncapi/studio/issues/80


src/reportWebVitals.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const reportWebVitals = (onPerfEntry?: ReportHandler) => {
+  if (onPerfEntry && onPerfEntry instanceof Function) {
+    import('web-vitals').then(({ getCLS, getFID, getFCP, getLCP, getTTFB }) => {
+      getCLS(onPerfEntry);
+      getFID(onPerfEntry);
+      getFCP(onPerfEntry);
+      getLCP(onPerfEntry);
+      getTTFB(onPerfEntry);
+    });
```
BOLT04, line 11: pretty cool @magicmatatjahu 💯 💯 
Will Studio have Google Analytics? Perhaps we could send this info to analytics too (https://www.npmjs.com/package/web-vitals#user-content-send-the-results-to-google-analytics), does it make sense to do so?
Just asking because I think I saw some discussion on getting analytics from the playground, so Studio could have it too 🙂

What do you think?

magicmatatjahu: The file was created with `create-react-app` script, so it was added automatically. in future we will go with this as well as google analytics, but not now. Right now it's more important to get the Studio up and running. Google analytics and other things later :)

PR #96, opened 2021-09-29
chore: init new version of Studio

magicmatatjahu:
**Description**

Init new version of studio:
- setup env with `create-react-app`
- create main content of new Studio: Editor based on `monaco-editor` and rendered html by `@asyncapi/react-component`

PR has content splitted from https://github.com/asyncapi/studio/pull/88 PR and it introduces only needed core functionality of studio - editor and react-component. The `Terminal`, `EditorDropdown` etc will be introduced in other PRs to reduce code to review. 

A large amount of code is related to package-lock and the bundled web-workers for the monaco editor - for `json`, `yaml` languages and the editor itself.

Test will be also added in the next PRs.

**Related issue(s)**
See also https://github.com/asyncapi/studio/issues/80


jonaslagoni (changes requested):
Starting off with the basics most glaring things 😆 

Something is off with the design, the editor is not 100% height and leaves a gap, see below:

billede

- If there is an error in the AsyncAPI file, the preview completely disappears and just leaves the editor.
- There are no real visible error messages, only a small one in the editor itself, is there any way we can make this more visible to the users?
- Do you not want the release pipeline for the initial versions? Otherwise, the package.json file needs some more configs.
-The menu to the left does not work at all.
- No icon for the website.

magicmatatjahu:
@jonaslagoni Thanks for review! Yeah I know about these problems and I should described why it looks that way. As I wrote in PR's description, PR itself has a code separate from #88 PR, which has about 2 times as much code and components. Some things look strange or don't work, because the code that makes them work is not currently in the PR, but will be added in the next ones to make as few changes as possible to the review, like:
- `If there is an error in the AsyncAPI file, the preview completely disappears and just leaves the editor.` - this is due to the fact that there is currently no "splitting" functionality added to the editor and template windows - when an error occurs, the editor flex is stretched to its full width - a strange bug due to the lack of logic that exists in #88 PR.
- `There are no real visible error messages, only a small one in the editor itself, is there any way we can make this more visible to the users?` - it will be visible in Terminal which will be allocated in this "blank" window under editor :)
- `Do you not want the release pipeline for the initial versions? Otherwise, the package.json file needs some more configs.` - the release pipeline will be fixed before first release. Please note that I will merge changes to the `spa-scratch` branch, so release will be done after merging to master, so after few PR in `spa-scratch`. Don't worry, I will handle that, but not in this PR. It's only to create "initial" architecture of new repo's content.
- `-The menu to the left does not work at all.` - it will be work when navigation of document (with messages, operations etc) will be introduced in seperate PR, so at the moment it only renders, but I can remove it and introduce sidebar with navigation if this is problem for you.
- `No icon for the website.` - in the next PR ok? I will remember about that :)

magicmatatjahu:
@jonaslagoni I also removed for you (😘 ) no needed code atm - it will be introduced with other PRs.

jonaslagoni (approved):
👍

magicmatatjahu:
@jonaslagoni Thanks!

PR #99, opened 2021-09-30
chore: add terminal

magicmatatjahu:
**Description**

- add `Terminal` component like in VSC
- add `Problems` tab in the `Terminal` component to show all errors from our ParserJS: it shows all errors with appropriate line

image

Related issue(s)
See also #80


src/components/Terminal/TerminalTabs.tsx
```diff
@@ -0,0 +1,95 @@
...
+  if (tabs.length === 0) {
+    return null;
+  }
+
+  return (
+    <div>
+      <div
+        className="flex flex-row justify-between items-center px-2 border-b border-gray-700 text-white uppercase font-bold text-xs cursor-pointer"
+        onClick={e => {
+          const clientRects = e.currentTarget.parentElement?.parentElement?.getClientRects()[0];
```
jonaslagoni, line 33: 🧐

jonaslagoni (commented):
Got a few comments:
- Can you please update the parser dependency, as `2.2` is not supported.
- I get a "problem" that the terminal does not catch, I assume it is just a warning? That the AsyncAPI version is not supported. Anything that needs to be changed now?
billede
 - You don't want to add a few tests, or are you planning to create them afterwards 😟 ?

src/components/Terminal/TerminalTabs.tsx
magicmatatjahu: Yes 😅  I need to get how many pixels away from the bottom of an element so that the terminal closes itself on click - I am taking a few pixel adjustments to improve the UX.

magicmatatjahu:
@jonaslagoni 
- `Can you please update the parser dependency, as 2.2 is not supported.` - all deps are bumped
- `I get a "problem" that the terminal does not catch, I assume it is just a warning? That the AsyncAPI version is not supported. Anything that needs to be changed now?` - strange, I can use `2.0.0` without any warning:

image

- `You don't want to add a few tests, or are you planning to create them afterwards 😟 ?` - I had a plan from beginning to create them afterwards :)

jonaslagoni:
>     * `I get a "problem" that the terminal does not catch, I assume it is just a warning? That the AsyncAPI version is not supported. Anything that needs to be changed now?` - strange, I can use `2.0.0` without any warning:

I could make it disappear as well, no idea how refreshed the window and it reappeared.

magicmatatjahu:
@jonaslagoni Ok, it should be fixed now :)

jonaslagoni (approved):
LGTM 🎉

magicmatatjahu:
@jonaslagoni Thanks!

PR #105, opened 2021-10-05
feat: show convert modal when user uses older versions of spec

magicmatatjahu:
**Description**

Similar logic like in the [playground](https://playground.asyncapi.io/):

- Show modal when user will use older version (less than `2.X.X` version) - always show
- Show modal when user will use non latest version - only one time per day. Information about showing is stored in the sessionStorage, so it will be also showed when user will start new session in the browser.
- Add unit tests

image

**Related issue(s)**
See also #80

### DIFF
--- a/src/components/Modals/ConvertToLatestModal.tsx
+++ b/src/components/Modals/ConvertToLatestModal.tsx
@@ -1,0 +1,76 @@
+import React, { useEffect, useState } from 'react';
+import toast from 'react-hot-toast';
+
+import { ConfirmModal } from './ConfirmModal';
+
+import { EditorService, SpecificationService } from '../../services';
+import state from '../../state';
+
+export const ConvertToLatestModal: React.FunctionComponent = () => {
+  const [show, setShow] = useState(false);
+  const specState = state.useSpecState();
+  const shouldOpenConvertModal = specState.shouldOpenConvertModal.get();
+
+  const latestVersion = SpecificationService.getLastVersion();
+
+  useEffect(() => {
+    shouldOpenConvertModal && setShow(true);
+  }, [shouldOpenConvertModal]);
+
+  const onSubmit = () => {
+    async function convert() {
+      try {
+        await EditorService.convertSpec();
+      } catch (e) {
+        throw e;
+      } finally {
+        specState.shouldOpenConvertModal.set(false);
+        setShow(false);
+      }
+    }
+
+    toast.promise(convert(), {
+      loading: 'Converting...',
+      success: (
+        <div>
+          <span className="block text-bold">
+            Document succesfully converted!
+          </span>
+        </div>
+      ),
+      error: (
+        <div>
+          <span className="block text-bold text-red-400">
+            Failed to convert document.
+          </span>
+        </div>
+      ),
+    });
+  };
+
+  return (
+    <ConfirmModal
+      title="Convert AsyncAPI document to latest version"
+      confirmText={`Convert to ${latestVersion}`}
+      confirmDisabled={false}
+      show={show}
+      onSubmit={onSubmit}
+    >
+      <div className="flex flex-col content-center justify-center text-center">
+        <p>
+          Your document is using an old version of AsyncAPI. The Studio only supports 2.0.0 and above. Convert your document to latest ({latestVersion}) version to continue.
+        </p>
+        <div className="mt-8">
+          <a
+            className="underline"
+            href={`https://www.asyncapi.com/blog/release-notes-${latestVersion}`}
+            target="_blank"
+            rel="nofollow noopener noreferrer"
+          >
+            See the release notes
+          </a>
+        </div>
+      </div>
+    </ConfirmModal>
+  );
+};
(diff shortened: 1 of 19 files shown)


magicmatatjahu:
@boyney123 Do you wanna make another review? 😅

src/services/specification.service.ts
```diff
@@ -104,6 +107,29 @@ export class SpecificationService {
     return false;
   }
 
+  static shouldInformAboutLatestVersion(
+    version: string,
+  ): boolean {
+    const oneDay = 24 * 60 * 60 * 1000; /* ms */
```
jonaslagoni, line 113: Not sure what I feel about this 😄 In theory if you said no once, I think it should not ask again, but discretely show it somewhere, that the document is not up to date 🤔 Thoughts?

src/components/Modals/ConvertToLatestModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <div className="mt-8">
+          <a
+            className="underline"
+            href={`https://www.asyncapi.com/blog/release-notes-${latestVersion}`}
+            target="_blank"
+            rel="nofollow noopener noreferrer"
+          >
+            See the release notes
+          </a>
+        </div>
```
jonaslagoni, line 72: Should we not show the release notes for each version inbetween? i.e. if you have a document with version `2.0.0` and we currently are in `2.2.0` it should show release notes for `2.1.0` and `2.2.0` 🤔

src/services/specification.service.ts
magicmatatjahu: I don't want to show it on every time, so it shows once a day (or once a session) - just like in playground. I can show information that there is a new version in the terminal, next to `Valid/Invalid` info, that the document does not use the latest version. What do you think?

jonaslagoni: The terminal part sounds like a good idea 👍 

I don't think it should re-ask each day though, once is enough, but that's just my preference 😄

src/components/Modals/ConvertToLatestModal.tsx
magicmatatjahu: It only shows release notes for latest, but it's a good idea to show all between current version and latest. I will propagate it. Thanks!

magicmatatjahu: Mock for `2.0.0` version (and also for older than `2.0.0` - we don't have release notes for `2.0.0` so I won't show any link):

image

Mock for `2.1.0` version:

image

jonaslagoni: This reminds me based on last PR - https://github.com/asyncapi/studio/pull/100

Should the user be able to select another version to convert to then the latest version? 🤔

src/services/specification.service.ts
magicmatatjahu: image

src/components/Modals/ConvertToLatestModal.tsx
magicmatatjahu: I can do this :)

src/services/specification.service.ts
jonaslagoni: Can you add a red negative sign, opposite of `valid` basically, and make it so you can click it to convert? 🤔

src/components/Modals/ConvertToLatestModal.tsx
magicmatatjahu: But... Between 2.0.0 and 2.2.0 there are only new things that help more than hinder. We could implement something like this when we will release 3.0.0, what do you think? I can't think of a use case if someone wanted a version smaller than 2.2.0 (atm).

src/services/specification.service.ts
jonaslagoni: Just want the `NOT LATEST` to be more visible to the user 🙂

src/components/Modals/ConvertToLatestModal.tsx
jonaslagoni: It makes sense yea, I cant think of a use-case either, however when you have the option in the other functionality to decide between versions, I would assume the same behavior where consistent here as well 🙂

magicmatatjahu: Ok, I will do it tomorrow :) Thanks for suggestions!

src/services/specification.service.ts
magicmatatjahu: Ok, I will do it tomorrow :) Thanks for suggestions! `NOT LATEST` will have warning sign.

magicmatatjahu:
@jonaslagoni Changes:

Info about not latest version in terminal:

image

After clicking in the `Not latest` button we will see modal to convert doc to latest version (without selection):

image

Modal that shows up to convert old (not supported) versions - cancel button is disabled:

image

Modal that shows up to convert old (but supported) versions:

image

jonaslagoni (approved):
LGTM 👍

magicmatatjahu:
@boyney123 No worries :)

PR #108, opened 2021-10-06
feat: add new functionalities to the editor dropdown

magicmatatjahu:
**Description**

Add new functionalities to the editor dropdown:

- possibility to convert JSON <-> YAML
- possibility to save given AsyncAPI doc to file
- possibility to save given AsyncAPI doc to file after converting JSON <-> YAML

image

**Related issue(s)**
See also #80

### DIFF
--- a/src/services/editor.service.ts
+++ b/src/services/editor.service.ts
@@ -1,4 +1,5 @@
 import * as monacoAPI from 'monaco-editor/esm/vs/editor/editor.api';
+import fileDownload from 'js-file-download';
 
 import { FormatService } from "./format.service";
 import { SpecificationService } from "./specification.service";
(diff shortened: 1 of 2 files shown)


src/services/editor.service.ts
```diff
@@ -101,4 +102,50 @@ export class EditorService {
       throw err;
     }
   }
+
+  static async convertToYaml() {
```
jonaslagoni, line 106: Why is it an `async` function? 🤔 I can see it is for the toaster, but there is no asynchronous code, is it really needed? 😄

src/services/editor.service.ts
```diff
@@ -101,4 +102,50 @@ export class EditorService {
...
+      console.error(err);
+      throw err;
+    }
+  }
+
+  private static fileName = 'asyncapi';
+
+  private static downloadFile(content: string, fileName: string) {
+    return fileDownload(content, fileName);
+  }
```
jonaslagoni, line 150: 🧐

jonaslagoni:
Sorry, thought the PR was ready, got pinged on slack 😄

src/services/editor.service.ts
magicmatatjahu: `fileName` is a base name, and then to it is added extension - see functions `saveJSON` etc :) There are any problems with shadowing etc.

src/services/editor.service.ts
magicmatatjahu: Yes, `async` is needed, because I wanna treat function as Promise to not block the main thread that renders the html - every state/actions in front should be asynchronous.

jonaslagoni: Fair 👍

magicmatatjahu:
@jonaslagoni Thanks!

PR #113, opened 2021-10-07
feat: support URL parameters

magicmatatjahu:
**Description**

Support URL parameters like in playground:
- `load` - load the external AsyncAPI spec via URL
- `url` - this same purpose as load
- `bse64` - load the AsyncAPI document saved in base64
- `readOnly` - show only rendered template. Allowed values are `true` and empty string

Example with `readOnly`:

image

**Related issue(s)**
Part of #80


magicmatatjahu:
@jonaslagoni If you have in mind linter errors, they're exactly fixed 😆

PR #178, opened 2021-11-17
feat: add splash (loading) screen

magicmatatjahu:
**Description**

- Improve UX by adding splash (loading) screen.
- Purge unnecessary Tailwind's classes by adding tailwind.css config - reduce final bundle size.


public/index.html
```diff
@@ -28,10 +28,35 @@
...
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/favicon-194x194.png" />
 
+    <link rel="stylesheet" type="text/css" href="css/preloader.css">
+
     <title>AsyncAPI Studio</title>
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
+
+    <div class="preloader">
```
boyney123, line 38: Just out of interest really, but how come the .css file was added for this preloader? Can most of it be done using tailwind stuff we have?

magicmatatjahu: both yes and no. It's hard to do tailwind animations, especially rotating wheel, but I will try to remove redundant classes. Also I'm using this rotating wheel in several places, so easier for me is creating separate css class for that.

magicmatatjahu: Ok, removed :)

public/index.html
```diff
@@ -32,6 +32,28 @@
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
+
+    <div id="preloader">
```
boyney123, line 36: Hey @magicmatatjahu 

Just wondering, how come you decided to add the reloader here inside the html vs inside the react component?

Was it a performance thing? You wanted to show the loader ASAP, whilst the app boots up?

My initial thought would be to do it in the app.jsx or somewhere, but yeah just wanted to know your thought process :)

magicmatatjahu: > You wanted to show the loader ASAP, whilst the app boots up?

Yes :) If we will add loader to the main component, then it will be rendered when script will be loaded and that it will show immidiatelly :)

boyney123: Yeah nice touch never thought of that before, loader looks great btw. Cool!

src/components/Editor/MonacoWrapper.tsx
```diff
@@ -52,7 +52,7 @@ export const MonacoWrapper: React.FunctionComponent<MonacoWrapperProps> = ({
   const onChange = debounce((v: string) => {
     EditorService.updateState({ content: v });
     SpecificationService.parseSpec(v);
-  }, 250);
```
boyney123, line 55: @magicmatatjahu any reason this was changed to 625? Was it to help with performance?

magicmatatjahu: I noticed that 250 msc is not that long time for garbarge collector to remove old references to old VDOM from React. those 625 are not that long time again but some references are able to remove themselves and free up memory faster :)

boyney123: Ah o wow, cool thanks for explaining :)

boyney123 (approved):

boyney123:
Nice work @magicmatatjahu , like the loader, looks great and gives a polished experince.

magicmatatjahu:
@mcturco Do you like the loading screen or you wanna give some insights? :)

mcturco:
This looks awesome @magicmatatjahu !! I'd say merge it! Definitely a huge improvement from how it currently loads in. Well done!

Issue #198, opened 2021-12-02
Templates section button does not trigger events for closing or opening previous tabs


smoya:
#### Describe the bug
Unlike the events visualizer button, clicking on the new templates button opens a new view that covers the whole studio (except the nav bar). When clicking then on any other button like the code one ![CleanShot 2021-12-02 at 12 31 41](https://user-images.githubusercontent.com/1083296/144414192-dcd35aa1-cfdb-49d3-9a52-bf363ced31e4.png), doesn't quit the examples view.

The only way to close it is to click again on the same examples button, which is not a 100% expected behavior because it is quite inconsistent with the rest.

#### How to Reproduce
See the following video:

https://user-images.githubusercontent.com/1083296/144412090-a9f801b4-171d-4848-b17a-ec54f90cb8fd.mp4

#### Expected behavior
Either examples view shows as a popup, or either just as a side view (as the events visualizer, code examples, etc), or we improve the navigation so on any click in any button will close the examples view.

magicmatatjahu:
> Either examples view shows as a popup, or either just as a side view (as the events visualizer, code examples, etc), or we improve the navigation so on any click in any button will close the examples view.

The question is which would be better? From a UX standpoint, I think modal, @mcturco @boyney123 any thoughts? 🤔

mcturco:
@magicmatatjahu Agreed! I think a modal for the new templates will work just fine. So then the modal should close if you click a different nav item on the left side, as @smoya pointed out, good catch!

magicmatatjahu:
Nope, no close by clicking on the nav item. It will be similar to normal modal like:

![image](https://user-images.githubusercontent.com/20404945/144632587-b736b3aa-2745-49dc-b44d-e2666b060acd.png)

But content will be templates :) And then you cannot click on the nav - you can try.

mcturco:
@magicmatatjahu Oh! Gotcha. Yes that works too!

boyney123:
Yeah modal sounds good to me! Nice catch @smoya

boyney123:
I can try and pick this up 👍

boyney123:
Think it's fixed in #201

magicmatatjahu:
Fixed in the https://github.com/asyncapi/studio/pull/201

PR #201, opened 2021-12-06
feat: move template selection into modal UI

boyney123:
**Description**

Related to issue #198

Moved the template selection into a modal now rather than it's own page.

To make this work I had some make some minor changes:

## Changes
- Changes to the UI to fit inside modal
- Allow modal to accept custom classes so I could make it wider by default (lots to render on this page, max-w-2xl was not big enough, looked weird).
- Removed a bug where two templates were getting rendered at the same time (currently in production)

## **UI: When user does not select anything**

image

## **UI: When user selects something**

image

Selecting a template then clicking "Create" will do as expected and render the new AsyncAPI file into the studio.

Let me know what you think @smoya , @magicmatatjahu @mcturco , hopefully this fixes the problem.


smoya:
Love it! Question: could we do a solid line instead of dash-style for the selected template so it becomes more visible?

src/components/Modals/NewFileModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { EditorService } from '../../services';
+import state from '../../state';
+
+export const NewFileModal: React.FunctionComponent = () => {
+  const sidebarState = state.useSidebarState();
+  const [selectedTemplate, setSelectedTemplate] = useState({ title: '', template: '' });
+
+  const handleTemplateClick = (template: any) => {
+    setSelectedTemplate(template);
+  };
```
magicmatatjahu, line 16: As I see, you don't make any logic here but only repass the arguments, so please remove that function and start using `setSelectedTemplate` function in places where you used the `handleTemplateClick` function.

src/components/Modals/NewFileModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+        </span>
+      </div>
+    );
+  };
+
+  const realLifeExamples = examples.filter((template) => template.type === 'real-example');
+  const templates = examples.filter((template) => template.type === 'protocol-example');
+
+  return (
+    <ConfirmModal containerClassName="sm:max-w-6xl" onCancel={() => sidebarState.panels.newFile.set(false)} title="AsyncAPI Templates - Start with our template examples" confirmText="Create" confirmDisabled={false} show={true} onSubmit={handleSubmit}>
```
magicmatatjahu, line 38:
```suggestion
    <ConfirmModal containerClassName="sm:max-w-6xl" onCancel={() => sidebarState.panels.newFile.set(false)} title="AsyncAPI Templates - Start with our template examples" confirmText="Use Template" confirmDisabled={false} show={true} onSubmit={handleSubmit}>
```

src/components/Modals/NewFileModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+    EditorService.updateState({ content: selectedTemplate.template, updateModel: true });
+    const panels = state.sidebar.panels;
+    panels.merge({
+      newFile: false,
+    });
+
+    toast.success(
+      <div>
+        <span className="block text-bold">
+          Succesfully created new file from {`"${selectedTemplate.title}"`} template.
```
magicmatatjahu, line 28:
```suggestion
          Succesfully reused the {`"${selectedTemplate.title}"`} template.
```

magicmatatjahu (changes requested):
First of all, awesome! Thanks for fast contribution! But I have two problems:

- we shouldn't make active nav item in the navigation, when someone click on that nav item. It doesn't look good for me. We should only open modal.
- we shouldn't indicates that we "create" new file by templates, but use template, atm we don't create new file but reuse templates, so please change all occurences of "creating file" to the "use" or "reuse" - I made suggestion for that :)

And also please propagate suggestion from @smoya

mcturco:
> could we do a solid line instead of dash-style for the selected template

to add to this suggestion by @smoya, I would also prefer that the inactive template card have a solid line as well. The light gray alone will be enough for the user to know that it is inactive. :)

boyney123:
Thanks @smoya , @mcturco @magicmatatjahu 

> to add to this suggestion by @smoya, I would also prefer that the inactive template card have a solid line as well. The light gray alone will be enough for the user to know that it is inactive. :)

Thanks @smoya and @mcturco , good ideas. That has been changed.

---

>we shouldn't make active nav item in the navigation, when someone click on that nav item. It doesn't look good for me. We should only open modal.

Thanks @magicmatatjahu , think I have fixed that returning `false` from Sidebar.tsx (is that alright?)

```js
 {
      name: 'newFile',
      state: () => false,
      icon: <VscNewFile className="w-5 h-5" />,
    },
```

> we shouldn't indicates that we "create" new file by templates, but use template, atm we don't create new file but reuse templates, so please change all occurences of "creating file" to the "use" or "reuse" - I made suggestion for that :)

Yeah good point mate. Changes made.

---

Let me know if you have any more thoughts folks 👍

mcturco:
Hey @boyney123! I am still seeing the dashed lines on the inactive templates. The solid border on the selected template looks great!! 😄  I was thinking that it would look more uniform if we kept the borders solid for all of them, but if we don't think that looks good we don't have to do that!!

mcturco:
Oh! Also I just noticed that the "beta" label is now black in the upper left where the logo is.

Screen Shot 2021-12-13 at 9 54 03 AM

Not sure if that was intentional, but just wanted to point it out either way. 😄

magicmatatjahu:
@mcturco Yeah, Fran created issue for that 😄  https://github.com/asyncapi/studio/issues/209

magicmatatjahu:
@boyney123 

> Thanks @magicmatatjahu , think I have fixed that returning false from Sidebar.tsx (is that alright?)

Yeah, it should works :)

boyney123:
> I am still seeing the dashed lines on the inactive templates

Sorry @mcturco I forgot to push up my changes 😅 , try again 🤞 🙏

magicmatatjahu (dismissed):
LGTM!

magicmatatjahu:
@boyney123 You should rebase your PR with master branch to resolve merge conflicts (`package-lock.json`) :)

smoya:
> @boyney123 You should rebase your PR with master branch to resolve merge conflicts (`package-lock.json`) :)

This is due to npm version mismatch. Just checkout `package-lock.json` from source (meaning should be exact same content). 

More about this: https://github.com/asyncapi/parser-js/issues/427

mcturco:
Awesome @boyney123!! Looks great 🎉

boyney123:
Thanks folks.

Thanks @smoya , yeah I have no idea what is going on with the lock file, as nothing has changed, I just reused the file from the master branch as suggested.

magicmatatjahu:
@boyney123 I bumped npm to v8 and from v7 there is a new version and structure of package-lock, Sergio created issue for that https://github.com/asyncapi/parser-js/issues/427 Please also fix the duplication of code - https://sonarcloud.io/component_measures?id=asyncapi_studio&metric=new_duplicated_lines_density&pullRequest=201&selected=asyncapi_studio%3Asrc%2Fcomponents%2FModals%2FNewFileModal.tsx&view=list - and I will accept :)

src/components/Modals/NewFileModal.tsx
```diff
@@ -0,0 +1,104 @@
...
+                    <div className="flex justify-between w-full">
+                      <span className={`block text-md  font-bold leading-0 group-hover:text-pink-600 ${textStyles} `}>{title}</span>
+                      {isSelected && <BsFillCheckCircleFill className="w-5 h-5 text-pink-600" />}
+                    </div>
+                    <span className="block text-sm text-gray-500 font-light mt-1 group-hover:text-gray-900">
+                      <Description />
+                    </span>
+                  </button>
+                );
+              })}
```
BOLT04, line 60: Hi @boyney123 🙂, I took a look at the Netlify preview, and this new modal looks amazing! I noticed SonarCloud found duplicate lines so maybe we could separate this logic into its own component, what do you think?

Something like this:
```ts
interface TemplateListItemProps {
  title: string;
  description: React.FunctionComponent;
  template: string;
}

const TemplateListItem: React.FunctionComponent<TemplateListItemProps> = ({
  title,
  description: Description,
  template,
  isSelected
}) => {
  const containerStyles = isSelected ? 'border-pink-500' : 'border-gray-200';
  const textStyles = isSelected ? 'text-pink-600' : 'text-gray-600';

  return (
    <button
      onClick={() => setSelectedTemplate({ title, template })}
      key={title}
      className={`group text-left flex flex-col cursor-pointer rounded-lg p-4 pb-6 border-2 hover:border-pink-500 ${containerStyles}`}
    >
      <div className="flex justify-between w-full">
        <span className={`block text-md font-bold leading-0 group-hover:text-pink-600 ${textStyles}`}>{title}</span>
        {isSelected && <BsFillCheckCircleFill className="w-5 h-5 text-pink-600" />}
      </div>
      <span className="block text-sm text-gray-500 font-light mt-1 group-hover:text-gray-900">
        <Description />
      </span>
    </button>
  );
};
```

Also, it seemed like the HTML used in the `realLifeExamples` case had an extra `border-gray-200`, and if it was selected it had these two classes:
imagem

boyney123: Just looking at this now, thank you @BOLT04 , I used your example and it worked fine (with a few tweaks!). Thanks for that.

Also the border duplication is removed, nice catch!

boyney123:
Fixed duplication thanks for the review @magicmatatjahu and @BOLT04 

Think it's ready for review again.

> I bumped npm to v8 and from v7 there is a new version and structure of package-lock (you probably use npm v6)

Also another node.... I upgraded to npm `8.3.2` which makes the lock file better, but it has added `"dev": true` to some packages, is this expected? Any idea @magicmatatjahu ?

magicmatatjahu:
@boyney123 That label is added to packages installed from `devDependencies` list :) 

https://stackoverflow.com/questions/49809490/what-is-dev-true-in-package-lock-json-for

src/components/Modals/NewFileModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+import toast from 'react-hot-toast';
+
+import { ConfirmModal } from './index';
+import examples from '../../examples';
+import { EditorService } from '../../services';
+import state from '../../state';
+
+interface TemplateListItemProps {
+  title: string;
+  description: React.FunctionComponent;
```
magicmatatjahu, line 12: I think that there should be better type to also "covers" ClassComponents.

```suggestion
  description: React.ComponentType;
```

magicmatatjahu (changes requested):
Awesome!

I found two bugs:
- when I don't have selected template I can click `Use Template` and it replaces my current AsyncAPI spec with empty content, you can test it :) We should disable the `Use Template` button when user doesn't select template.
- another is described in the separate comment.  

Just these two things for me aren't acceptable, rest is ok :)

magicmatatjahu:
@boyney123 I applied my suggestions :) Could you check if you accept it and we can merge it :)

boyney123:
Thank you @magicmatatjahu yeah looks OK to me, just checked your changes and happy with them.

I can't approve the PR as I'm the author so guess this comment is the approval 🤣👍

magicmatatjahu (approved):
LGTM! 😆

magicmatatjahu:
@mcturco Do you accept current solution?

mcturco:
@magicmatatjahu yes, I actually just tested it and was about to write a comment that it looks good to me!! 🎉

PR #211, opened 2021-12-13
fix: some security issues

fmvilas:
This is fixing some moderate security issues. It's the result of running `npm audit fix`.


magicmatatjahu (approved):
LGTM! I don't see any discrepancies :)

Issue #193, opened 2021-11-30
Add tooltips to navigation and icons to new examples screen


mcturco:
#### Reason/Context
I am wondering if the icons on the navigation properly explain what the window is that corresponds to it. I suppose users who are frequent on the studio might know from learning the system, but in order to make it easier for new users to quickly understand the system, I am proposing that we add tooltips on hover of each navigation item so there are proper labels for each window pane.

I think in addition to this, there might be a way where we can make the navigation items that are active have a more obvious style than just the white color and border on the left side. I feel that sometimes the white border on the left can get lost depending on where you have your window positioned on your computer, especially because someone may have the window bleeding off the left side and might miss the white border entirely.

I can create a mock up for these ideas soon and share it on this issue, but just wanted to add in this issue ASAP.

magicmatatjahu:
@mcturco Thanks for ideas, I like them! Regarding to the color of the active nav items, we can change left white border to the pink (this color of `studio` title) and also make more contrast, I'm thinking of brightening up the background of nav item. I will wait anyway for your mocks :)

boyney123:
`I am proposing that we add tooltips on hover of each navigation item so there are proper labels for each window pane.`

💯 , think maybe we can also add #199 here too?

PR #215, opened 2021-12-16
feat: added tooltips for navigation bar (#193)

boyney123:
**Description**

Following #193, I added some tooltips to the navigation bar.

Let me know what you think @mcturco @magicmatatjahu


src/components/Sidebar.tsx
```diff
@@ -116,6 +117,7 @@ export const Sidebar: React.FunctionComponent<SidebarProps> = () => {
       <div className="flex flex-col">
         <SettingsModal />
       </div>
+      <ReactTooltip effect='solid' backgroundColor='#374151' arrowColor='#374151' className="text-xs"  />
```
magicmatatjahu, line 120: So if I wanna add tooltip in another part of Studio, e.g. in Editor, then I should copy that component and use `data-tip` attributes? 🤔  I think that component that wraps given component and add tooltip will be better, something like:

```tsx
<Tooltip content={...}>
  //...some component
</Tooltip>
```

is there any another tooltip component in the OpenSource?

magicmatatjahu (changes requested):
I have a problem where exactly tooltip is showed, check this:

image

image

as you can see if I move the cursor in the nav item, but not at the icon I don't see any tooltip. It should be integrated not with icon, but with whole nav item :)

mcturco:
Thanks @boyney123 for adding this!! 🎉   However, I think if we made the background of the labels fully opaque it would look a lot cleaner. I think it might be a little distracting to see things peeking through behind the labels as the text size is already rather small. What do you think?

mcturco:
@boyney123 @magicmatatjahu 

Hey guys! Was cleaning out my inbox and realized that this PR has yet to be merged. I think this is still a great feature that we should consider revisiting and merging. I think all we should do is make the tool tips fully opaque instead of transparent and integrate them with the entire button and not just the icon. Then we should definitely ship it! What do you guys think?

magicmatatjahu:
Tooltips added in the https://github.com/asyncapi/studio/pull/434 PR 

@boyney123 I hope that you don't have a problem that I'm closing this PR :)

PR #222, opened 2022-01-04
fix: update react-component to next-27 version

magicmatatjahu:
**Description**

Auto bumping script doesn't work -> https://github.com/asyncapi/asyncapi-react/runs/4700884362?check_suite_focus=true so I bump that dep manually.

**Related issue(s)**
Resolves https://github.com/asyncapi/studio/issues/221


magicmatatjahu:
@boyney123 Thanks! 🚀

Issue #86, opened 2021-09-22
Create (server) API for Studio


magicmatatjahu:
In https://github.com/asyncapi/studio/issues/80 proposal was proposed to create a (server) API that would later be used by the Studio for things that cannot be done in the browser. Most of our tools are isomorphic, but e.g. generator still need to run a NodeJS environment. Issue is opened to discuss how this API should look like, what it should offer, in which technology/framework it should be written and in which way it should be distributed.

@jonaslagoni @derberg @fmvilas @smoya You should be interested in this topic :)
@BOLT04 If you want also add comment, feel free, but I'm not forcing anything.

Closed issues (related to old implementation of the Studio) but are related to the server API and should be consider to implement:
- https://github.com/asyncapi/studio/issues/9
- https://github.com/asyncapi/studio/issues/12
- https://github.com/asyncapi/studio/issues/8 - plugins for server API
- https://github.com/asyncapi/studio/issues/14 - Google Analitycs plugin - mainly in front-end app, but here it can be also used.

smoya:
Do we have a list of those actions the server should perform, e.g. generation of templates?
In terms of infrastructure estimation costs, what is the estimated traffic for that API? Would it be similar to current playground?

magicmatatjahu:
@smoya 

> Do we have a list of those actions the server should perform, e.g. generation of templates?

At the moment in plans is implement things from playground, so generating and downloaded result of templates. In future we can implement (as @jonaslagoni suggest) the model generation, optimizations etc, but a lot of things we can perform in browser, because our libraries are isomorphic, so the most important thing is the generation of templates. The list is open and we can add to it whatever we/community need.

> In terms of infrastructure estimation costs, what is the estimated traffic for that API? Would it be similar to current playground?

It depends in which way we will go - in which form we will distribute that API - because we have several options:
- treat API as separate package that anyone can use to set up a server at their servers.
- publish api with (e.g.) `api.asyncapi.com` domain, that everyone will be able to request.
- merging of two above - everyone can be able to set up new server, but also be able use server (by free) hosted by AsyncAPI Initiative

Of course, for first option, the costs are 0, for second it depends on traffic 😅  So at the moment I don't know. We should check what traffic we have on https://playground.asyncapi.io/ and by this estimate costs. But we must also remember that the community will grow, so costs also. We can also limit hourly/IP requests to reduce costs etc.

BOLT04:
[...]

PR #229, opened 2022-01-11
feat: add integration with AsyncAPI Server API

magicmatatjahu:
**Description**

- add `ServerAPI` service
- add `GeneratorModal`:

  image
  
  We can open it by clicking `Generate template` dropdown option:
 
  image

- generate all available templates in our organization:
   - '@asyncapi/dotnet-nats-template'
   - '@asyncapi/go-watermill-template'
   - '@asyncapi/java-spring-cloud-stream-template'
   - '@asyncapi/java-spring-template'
   - '@asyncapi/java-template'
   - '@asyncapi/nodejs-template'
   - '@asyncapi/nodejs-ws-template'
   - '@asyncapi/python-paho-template'
   - '@asyncapi/ts-nats-template'
- add script to generate JSON Schemas from template parameters (for each template) which we use to create form for passing parameters. In future it will be replaced by https://github.com/asyncapi/community/issues/249 solution

**Related issue(s)**
Part of https://github.com/asyncapi/studio/issues/86


jonaslagoni:
Just a quick comment, I don't feel the `generate template` reflects what it does 🤔 You don't generate a template, you generate with a template. 

However, you cant say `generate code` as that's not entirely accurate 😅 `generate stuff`, `generate code/docs` hmm, @alequetzalli maybe you have a suggestion for better wording 🤔 ?

alequetzalli:
> Just a quick comment, I don't feel the `generate template` reflects what it does 🤔 You don't generate a template, you generate with a template.
> 
> However, you cant say `generate code` as that's not entirely accurate 😅 `generate stuff`, `generate code/docs` hmm, @alequetzalli maybe you have a suggestion for better wording 🤔 ?

I see, in that case @jonaslagoni and @magicmatatjahu, how about the following idea? 

Where is currently reads: **Generate template based on AsyncAPI Document**, how about changing it to `Generate code/docs with a template`.

Then underneath where it simply says **Template**, how about adding description text to it that reads `Your template will be based on your AsyncAPI Document.` 

Screen Shot 2022-01-21 at 2 32 17 PM

magicmatatjahu:
@alequetzalli 

> Where is currently reads: Generate template based on AsyncAPI Document, how about changing it to Generate code/docs with a template.
>
> Then underneath where it simply says Template, how about adding description text to it that reads Your template will be based on your AsyncAPI Document.

Great! I'll add that :) However, what about node in dropdown? `Generate template`, or `Generate docs/code` as @jonaslagoni suggested? WDYT?

image

alequetzalli:
> `Generate docs/code`

@magicmatatjahu  Ah! I for that drop down I would do what @jonaslagoni suggested too and replace it for `Generate docs/code`.

Screen Shot 2022-01-25 at 8 42 01 PM

boyney123:
@magicmatatjahu looking good mate!

I have a few questions, ideas and feedback around it, wondered what you thought?

---

### Template Selection Naming

It currently renders the package name (from what I can tell), I wonder if there is a more user-friendly way of showing this, without having to understand the package names etc?

Here is what it renders now 👇
image

Some Ideas 👇 (not all there, but some)

image

Think this might give more context for the user what they are doing?

---

### Hide params that are not required (with the option to show)

When you select a template the user is shown the params for that template, I'm assuming some of these are optional ones? Is it possible to have only the required ones shown and "Advanced Options" for the optional ones. I think this could simplify the UX for people? Just a thought!

image

---

### Link to Package Docs when clicking a Template

Coming at some of these templates new, it's hard to understand what the options mean even tho the description is there. I wonder if a link to the GitHub repo/docs for the templates could help people gain details if they need it?

---

These are just some thoughts after the initial look, wonder what you all think?

magicmatatjahu:
@boyney123 Thanks very much for that feedback! I propagated your ideas to code. Could you check? :)

alequetzalli (dismissed):
🚢🚢🚢

boyney123:
Nice one @magicmatatjahu !

Dropdown is great with the additional context!

Not sure if it's just me but the `Advanced Options` placement of the button feels off to me? wonder if there is somewhere else we could put it?

image

Maybe something like this?

image

What do you think @magicmatatjahu? Just an idea 🤔

magicmatatjahu:
@boyney123 Ok I propagated your suggestions :) Could you check again?

magicmatatjahu:
@mcturco @boyney123 Could you check final (I hope) implementation? 😄

mcturco:
@magicmatatjahu Hi 👋 

It could just be me & my lack of spec knowledge, but the Advanced options in the modal for the HTML website seem a bit vague (I haven't looked through the options for the other outputs) 

Screen Shot 2022-02-14 at 1 55 02 PM

I am wondering if we should add placeholder text to the inputs so there is an example of what parameter to use.

`sidebarOrganization` to me should be a set of two radio buttons, since there are only 2 options. I foresee a lot of user error with this since it is a text input with only 2 strings you can add in there.

Maybe in `baseHref` there should be some placeholder text like `/example-url/` so the user is aware of what syntax to use.

`version` - is this option really necessary? Couldn't they just adjust the version in the studio editor before they click `Generate code/docs`? I am just wondering how many use cases this option offers.

`singleFile` - the verbiage in the description text here is confusing. Perhaps it should say `Enable this option to generate a single HTML file with scripts and styles added within`

`outFileName` - I think we could add some placeholder text here that says `example.html` and wondering if the .html part could be added automatically in the visual logic of the input? So the user should be able to just type the file name without having to enter .html but they will still visually see the file extension within the input

`pdf` - Tbh I am not really a fan of the options being presented in their schema name, and it is specifically bothering me for this parameter. IMO it should say `Generate PDF` instead of just `pdf`. Other than that, I think we can update the verbiage to match the other true/false toggle that I commented on above: `Enable this option to include a PDF version of your documentation website`

----------

Just took a look at some of the other advanced options for different outputs... WOW there are a lot of options under Java Spring Project. Are these all necessary options? Are there any options that can get combined or show/hide based on server configuration? I am seeing `Only for Kafka` in a couple of places and `Only for MQTT` in some other places.... could we possibly only show these options if the server in the AsyncAPI file is set to that specific server?

----------

Last note: I am seeing this as an "error" message in one of the output options. `AsyncAPI document doesn't have at least one server with supported protocols. For the selected generation, these are supported: ws`

Screen Shot 2022-02-14 at 2 09 33 PM

Shouldn't we properly spell out the protocol so that the user knows what to correct in their AsyncAPI file to make this error disappear? `ws` --> `WebSocket` OR maybe we should be more human about it 😄  We could say: `Configure your AsyncAPI document server protocol to WebSocket to generate a NodeJS WebSocket Project`

mcturco:
> the way it looks is due to a certain problem we're currently having in the https://github.com/asyncapi/generator side.

@magicmatatjahu gotcha. I understand now that the Generator is the source for the options to render output, thank you for explaining that! Hmm I still think that we should consider the notes I left even if we have to do it in a separate PR for the  generator. I just feel like the way the UX is now in the studio makes it hard for the user to know what to do to implement the advanced options, but we can of course do a usability test to see if thats just my opinion or if people will actually have a problem with how it is now.

Sorry if I misunderstood what I was supposed to be reviewing 😅

magicmatatjahu:
No worries :) Good that you did this review because it shows that actually this UX is not working and it needs to be improved on the generator side to render the form better. However, I would like you to do a review if the UX itself (not taking into account those problems we have with parameters) is ok, so thank you very much! :) I will try tomorrow to implement this mapping `ws` -> `WebSocket` etc.

mcturco:
@magicmatatjahu sure, no problem. But yeah I think it definitely is cool to see the generator working within Studio 😄  I think the option to download code is relatively easy to use (if we don't think about the potential issues that I brought up, haha)

I say we should ship! 🚢

mcturco (approved):

PR #238, opened 2022-01-19
fix: add support for avro and openapi formats

### LINKED ISSUE #237: N[String(...)] is not a function on studio.asyncapi.com
#### Describe the bug
If I copy this snippet from https://github.com/asyncapi/avro-schema-parser
```
asyncapi: 2.2.0
info:
  title: Example with Avro
  version: 0.1.0
channels:
  example:
    publish:
      message:
        schemaFormat: 'application/vnd.apache.avro;version=1.9.0'
        payload: # The following is an Avro schema in YAML format (JSON format is also supported)
          type: record
          name: User
          namespace: com.company
          doc: User information
          fields:
            - name: displayName
              type: string
            - name: email
              type: string
            - name: age
              type: int
```
into https://studio.asyncapi.com/ I get error `N[String(...)] is not a function`.
If I download the studio locally and past the same snippet, it works.

#### How to Reproduce
see above

#### Expected behavior
No error

magicmatatjahu:
**Description**

Add support for avro and openapi formats.

Why now? Because it is the first time when someone used `avro` schema in the Studio and got an error.... - #237  

The issue mentions that locally it works, but on https://studio.asyncapi.com it does not. This is because of how webpack works in production mode and in development mode. In development mode webpack "sees" all the packages that we use in all parts of the studio and for example `react-component` underneath "adds" support for avro and openapi. Due to the fact that Studio uses a smaller version of react-component without parser (parser is used in studio, so it doesn't need it), after building the application, where webpack "removes" any unused code, support for `avro` and `openapi` is forgotten, so we have to add support for `avro` and `openapi` again.

I don't want to release it as `feat:` because it's a bug, we should support it from beginning but we forgot to check that.

**Related issue(s)**
Fixes #237

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,9 @@
       "version": "0.5.5",
       "license": "Apache-2.0",
       "dependencies": {
+        "@asyncapi/avro-schema-parser": "^1.0.1",
         "@asyncapi/converter": "^0.6.1",
+        "@asyncapi/openapi-schema-parser": "^2.0.1",
         "@asyncapi/parser": "^1.12.0",
         "@asyncapi/react-component": "^1.0.0-next.31",
         "@asyncapi/specs": "^2.12.0",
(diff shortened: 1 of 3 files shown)


dweber019:
I can confirm this works now for me with avro.

magicmatatjahu:
@boyney123 

> Are there any other parsers we want to add whilst we are here? Not sure if we have any others? Maybe you know?

We have also https://github.com/asyncapi/raml-dt-schema-parser but that parser only works in the NodeJS env so we cannot add it in the studio. We should think how to parse spec with raml by local `server-api` but not now :)

boyney123:
> We have also asyncapi/raml-dt-schema-parser but that parser only works in the NodeJS env so we cannot add it in the studio. We should think how to parse spec with raml by local server-api but not now :)

Ok thanks for the info @magicmatatjahu

boyney123 (approved):

PR #240, opened 2022-01-20
feat: create Docker image

### LINKED ISSUE #30: DockerHub Image
#### Reason/Context
It would be beneficial to developers to be able to run this locally and internally within organisations.

- How will this change help?
Pushing the docker image for every new release to docker hub

- What is the motivation? 
Swagger Editor also provides this https://hub.docker.com/r/swaggerapi/swagger-editor/

#### Description

On release builds the versioned docker image would be built and pushed to docker hub
no

- How could it be implemented/designed?
as part of the GibHub build

fgreinacher:
Fixes #30

Took some inspiration from https://github.com/asyncapi/server-api/pull/24

🛠 with ❤ at Siemens

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ npm install
 npm start
 ```
 
-and then go to [http://localhost:3000]().
+and then go to [http://localhost:3000](http://localhost:3000).
 
 ## Using it via Docker
 
@@ -56,7 +56,10 @@ Run:
 docker run -it -p 8000:80 asyncapi/studio
 ```
 
-and then go to [http://localhost:8000]().
+and then go to [http://localhost:8000](http://localhost:8000).
+
+The `asyncapi/studio` image is based on the official `nginx` image.
+Please refer to the [Nginx documentation](https://registry.hub.docker.com/_/nginx/) to learn how to e.g. pass a custom `nginx` configuration or plug in additional volumes.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -48,6 +48,16 @@ npm start
 
 and then go to [http://localhost:3000]().
 
+## Using it via Docker
+
+Run:
+
+```bash
+docker run -it -p 8000:80 asyncapi/studio
+```
+
+and then go to [http://localhost:8000]().
+
 ## Development
 
 1. Setup project by installing dependencies `npm install`
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+.git/
+.github/
+node_modules/
+build/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
-FROM node:16 as build
+FROM docker.io/library/node:16.13.2 as build
 COPY ./ ./
 RUN npm install && npm run build
 
-FROM nginx:1.21.5
+FROM docker.io/library/nginx:1.21.5-alpine
 COPY --from=build /build /usr/share/nginx/html/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM node:16 as build
+COPY ./ ./
+RUN npm install && npm run build
+
+FROM nginx:1.21.5
+COPY --from=build /build /usr/share/nginx/html/
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@asyncapi/studio",
-  "version": "0.5.5",
+  "version": "0.5.6",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@asyncapi/studio",
-      "version": "0.5.5",
+      "version": "0.5.6",
       "license": "Apache-2.0",
       "dependencies": {
         "@asyncapi/avro-schema-parser": "^1.0.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "@headlessui/react": "1.4.1",
         "@hookstate/core": "^3.0.12",
         "@monaco-editor/react": "^4.2.2",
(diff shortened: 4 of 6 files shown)


fgreinacher:
Verified in Gitpod:

image

fgreinacher:
Not sure if I should ping someone for review: @fmvilas @boyney123 @mcturco @magicmatatjahu Could one of you have a look?

magicmatatjahu:
@fgreinacher Hi! Don't worry, I will make review today. I saw that PR yesterday and that's what I wanted to do today :)

magicmatatjahu (commented):
@fgreinacher Hello! Thanks for your contribution! 

Dockerfile can be, but I wonder if we should rather go for a pure alpine image as base image - I mean the second image, the final one with nginx and the bundle - something like https://github.com/kyma-project/busola/blob/main/Dockerfile#L27 (please don't bother with a few bundles, certs etc but it's an idea to build an image on pure alpine). Then we would be able to define configurations for nginx. What do you think about it?

And of course please add a short documentation in the readme on how to run such an image (something like https://github.com/asyncapi/generator#cli-usage-with-docker). I forgot to add this in `server-api`, but it should be here too :) 

And sorry for the late review :)

fgreinacher:
> Dockerfile can be, but I wonder if we should rather go for a pure alpine image as base image - I mean the second image, the final one with nginx and the bundle - something like https://github.com/kyma-project/busola/blob/main/Dockerfile#L27 (please don't bother with a few bundles, certs etc but it's an idea to build an image on pure alpine). Then we would be able to define configurations for nginx. What do you think about it?

What do you think about using the `-alpine` flavor of the nginx image? It's [quite minimal](https://github.com/nginxinc/docker-nginx/blob/master/Dockerfile-alpine.template) and reduces stuff that needs to be maintained here. I can of course also switch to `alpine` and install nginx manually if you want.

> And of course please add a short documentation in the readme on how to run such an image (something like https://github.com/asyncapi/generator#cli-usage-with-docker). I forgot to add this in `server-api`, but it should be here too :)

Yes, of course!

> And sorry for the late review :)

No worries, it was actually pretty quick 😀

fgreinacher:
@magicmatatjahu I just pushed https://github.com/asyncapi/studio/pull/240/commits/dae2d14fc8794271ab88b395678dd511d62b7e18 where I switched to the Alpine-based Nginx image. Let me know what you think 😀

README.md
```diff
@@ -48,6 +48,16 @@ npm start
...
 
+## Using it via Docker
+
+Run:
+
+```bash
+docker run -it -p 8000:80 asyncapi/studio
+```
+
+and then go to [http://localhost:8000]().
```
magicmatatjahu, line 59: We have bug here 😅 

```suggestion
and then go to [http://localhost:8000](http://localhost:8000).
```

Could you also fix link (in the 49 line) in the dev section? I don't know that we have broken link here:

```
and then go to [http://localhost:3000](http://localhost:3000).
```

magicmatatjahu: Also, could you point in this section to the Docker Image docs (https://registry.hub.docker.com/_/nginx/)? People should know about customization of this image, how to pass custom nginx config or add more volumes, e.g:

```md
To learn how to e.g. pass a custom `nginx` configuration or plug in additional volumes please refer to the [documentation of the `nginx` docker image](https://registry.hub.docker.com/_/nginx/). 
```

magicmatatjahu (commented):
Hi! Thanks for new commits. I have two important comments :)

README.md
fgreinacher: Oh sorry, I blindly copy-pasted 😬

Please see https://github.com/asyncapi/studio/pull/240/commits/0730b3c9946865d2f4a99f958103dade0c9fc658.

magicmatatjahu: Fast! 😄  Thanks very much, everything is ok, I will check again locally that image and we can merge it! :)

magicmatatjahu: @fgreinacher We missed about most important part 😅  Could you pass that scripts https://github.com/asyncapi/server-api/blob/master/package.json#L38-L39 to the package.json `scripts`? Of course please change the `server-api` to the `studio`. `semantic-release-docker` plugin needs builded image, it doesn't built it itself.

magicmatatjahu:
@fgreinacher We missed about most important part 😅 Could you pass that scripts https://github.com/asyncapi/server-api/blob/master/package.json#L38-L39 to the package.json scripts? Of course please change the server-api to the studio. semantic-release-docker plugin needs builded image, it doesn't built it itself.

magicmatatjahu:
@fgreinacher Ok, docker works in my local machine, only problem I found that sometimes it throw stackoverflow exception and I need to increase max ram for Docker, but it's related to the local config, not to docker image, so please add that npm scripts and we can merge :)

fgreinacher:
> @fgreinacher We missed about most important part 😅 Could you pass that scripts https://github.com/asyncapi/server-api/blob/master/package.json#L38-L39 to the package.json scripts? Of course please change the server-api to the studio. semantic-release-docker plugin needs builded image, it doesn't built it itself.

@magicmatatjahu Oops, I assumed that this "just" works 😀 Fixed with https://github.com/asyncapi/studio/pull/240/commits/ce5bc4d2537d0112004ee95861fed872c5971d48.

magicmatatjahu (approved):
LGTM! 🚀

magicmatatjahu:
@fgreinacher Thanks for that amazing contribution! Docker is deployed and it works like a charm! https://hub.docker.com/r/asyncapi/studio/tags

fgreinacher:
> @fgreinacher Thanks for that amazing contribution! Docker is deployed and it works like a charm! https://hub.docker.com/r/asyncapi/studio/tags

Very cool! Thanks for the flawless contributor experience!

PR #248, opened 2022-01-26
feat: add support for configuring the base URL

### LINKED ISSUE #249: Run AsyncAPI Studio at subpath
#### Reason/Context

In some hosting scenarios (e.g. Docker Compose, Kubernetes) the container might not be exposed at the root path of the host.

AsyncAPI Studio does not handle this well at the moment:

image

I think it would be great to support this so that people can easily host the Studio in whatever environment they like.

fgreinacher:
**Description**

See the related issue for context.

This PR introduces the following changes:

- Sets `PUBLIC_URL` to a placeholder value during Docker build so `create-react-app` injects the placeholder value in generated assets (see https://create-react-app.dev/docs/advanced-configuration/ and https://create-react-app.dev/docs/deployment/#building-for-relative-paths)
- When the container is started replaces the placeholder value by the configured base URL. If no base URL is configured it will assume being served from root.
- Adapt some places where the `PUBLIC_URL` variable has not been used up to now (see https://create-react-app.dev/docs/using-the-public-folder/) 

**Related issue(s)**

Closes #249

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,32 @@
-FROM docker.io/library/node:16.13.2 as build
-COPY ./ ./
-RUN npm install && npm run build
+# Use a UUID as placeholder value to have a unique string to replace. 
+ARG BASE_URL_PLACEHOLDER=189b303e-37a0-4f6f-8c0a-50333bc3c36e
 
-FROM docker.io/library/nginx:1.21.5-alpine
+FROM docker.io/library/node:16.13.2 as build
+
+ARG BASE_URL_PLACEHOLDER
+
+COPY package.json package-lock.json ./
+RUN npm install --frozen-lockfile
+
+COPY ./ ./
+# Set the React PUBLIC_URL to our placeholder value so that
+# that it can easily be replaced with the actual base URL
+# in the entrypoint script below.
+RUN PUBLIC_URL=${BASE_URL_PLACEHOLDER} npm run build
+
+FROM docker.io/library/nginx:1.21.5-alpine as runtime
+
+ARG BASE_URL_PLACEHOLDER
+# The base Nginx image automatically executes all shell scripts 
+# within the /docker-entrypoint.d/ directory ("entrypoint scripts")
+# when the container is started. See the relevant logic at
+# https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/docker-entrypoint.sh#L16.
+ARG ENTRYPOINT_SCRIPT=/docker-entrypoint.d/set-public-url.sh
+
 COPY --from=build /build /usr/share/nginx/html/
+# Add an entrypoint script that replaces all occurences of the 
+# placeholder value by the configured base URL. If no base URL
+# is configured we assume the application is running at '/'.
+RUN echo "find /usr/share/nginx/html/ -type f -print0 | xargs -0 sed -i \"s|${BASE_URL_PLACEHOLDER}|\${BASE_URL}|g\"" > $ENTRYPOINT_SCRIPT && chmod +x $ENTRYPOINT_SCRIPT
+
+FROM runtime
--- a/README.md
+++ b/README.md
@@ -62,6 +62,13 @@ and then go to [http://localhost:8000](http://localhost:8000).
 The `asyncapi/studio` image is based on the official `nginx` image.
 Please refer to the [Nginx documentation](https://registry.hub.docker.com/_/nginx/) to learn how to e.g. pass a custom `nginx` configuration or plug in additional volumes.
 
+In some hosting scenarios (e.g. Docker Compose, Kubernetes) the container might not be exposed at the root path of the host.
+Set the environment variable `BASE_URL` to let AsyncAPI Studio know from where to resolve static assets:
+
+```bash
+docker run -it -p 8000:80 -e BASE_URL=/a/custom/path asyncapi/studio
+```
+
 ## Development
 
 1. Setup project by installing dependencies `npm install`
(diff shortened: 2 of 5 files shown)


src/index.tsx
```diff
@@ -10,14 +10,14 @@ window.MonacoEnvironment = window.MonacoEnvironment || {
...
     }
     // for yaml worker
     if (label === 'yaml' || label === 'yml') {
-      return '../../js/monaco/yaml.worker.bundle.js';
+      return './js/monaco/yaml.worker.bundle.js';
     }
     // for core editor worker
-    return '../../js/monaco/editor.worker.bundle.js';
+    return './js/monaco/editor.worker.bundle.js';
   },
```
fgreinacher, line 21: I assume the former paths just worked by accident because `..` does nothing when already at the root path.

fgreinacher:
@magicmatatjahu We noticed this after deploying the new Docker image from #240 to our Kubernetes cluster. Maybe you can have a look again?

magicmatatjahu:
@fgreinacher 

> @magicmatatjahu We noticed this after deploying the new Docker image from #240 to our Kubernetes cluster. Maybe you can have a look again?

Hi! No problem, I hope that I will make review today! I already made 4/5 review today and nexts wait for me 😅

src/index.tsx
magicmatatjahu: Right, if we load bundled JS file in the `index.html` then we can use path from root. In the preview I see that it works, but there is a question: why does it also work in current deploy on https://studio.asyncapi.com where we have that old path? 🤔

Dockerfile
```diff
@@ -1,6 +1,28 @@
+# Use a UUID as placeholder value to have a unique string to replace. 
+ARG BASE_URL_PLACEHOLDER=189b303e-37a0-4f6f-8c0a-50333bc3c36e
```
magicmatatjahu, line 2: I'm honestly not an expert when it comes to Docker, but why is `BASE_URL_PLACEHOLDER` defined three times? Isn't it better to define these arguments once at the top like:

```
ARG BASE_URL_PLACEHOLDER=189b303e-37a0-4f6f-8c0a-50333bc3c36e
ARG ENTRYPOINT_SCRIPT=/docker-entrypoint.d/set-public-url.sh
```

and remove other occurrences with `ARG`?

Dockerfile
```diff
@@ -1,6 +1,28 @@
...
+
 FROM docker.io/library/node:16.13.2 as build
+
+ARG BASE_URL_PLACEHOLDER
+
+COPY package.json package-lock.json ./
+RUN npm install --frozen-lockfile
+
 COPY ./ ./
```
magicmatatjahu, line 11: Why these commands are split? What we have benefit here?

Dockerfile
```diff
@@ -1,6 +1,28 @@
...
-RUN npm install && npm run build
+# Set the React PUBLIC_URL to our placeholder value so that
+# that it can easily be replaced with the actual base URL
+# in the entrypoint script below.
+RUN PUBLIC_URL=${BASE_URL_PLACEHOLDER} npm run build
+
+FROM docker.io/library/nginx:1.21.5-alpine as runtime
+
+ARG BASE_URL_PLACEHOLDER
+ARG ENTRYPOINT_SCRIPT=/docker-entrypoint.d/set-public-url.sh
```
magicmatatjahu, line 20: Could we add info about `docker-entrypoint.d`? Tbh I had to read about it, because I didn't know about that 😅  You know, for future contributors to help them understand that step.

README.md
```diff
@@ -62,6 +62,13 @@ and then go to [http://localhost:8000](http://localhost:8000).
 The `asyncapi/studio` image is based on the official `nginx` image.
 Please refer to the [Nginx documentation](https://registry.hub.docker.com/_/nginx/) to learn how to e.g. pass a custom `nginx` configuration or plug in additional volumes.
 
+In some hosting scenarios (e.g. Docker Compose, Kubernetes) the container might not be exposed at the root path of the host.
+Set the environment variable `BASE_URL` to let AsyncAPI Studio know from where to resolve static assets:
```
magicmatatjahu, line 66: I guess that you tested it on your environment and everything works? I tested it locally and base path `/` works and also preview works so for me it's most important to not broke https://studio.asyncapi.com. In your case with custom `BASE_URL` does it work like should? I don't have a way to test this on e.g. kubernetes so I rely on you here :)

src/index.tsx
magicmatatjahu: Also I wonder if we should use it here `process.env.PUBLIC_URL` to be more safe?

magicmatatjahu (changes requested):
@fgreinacher Thanks for your contribution! :) We are very thankful but I have a few comments, please refer to them :)

Dockerfile
fgreinacher: The variable value can be defined centrally, but it's unfortunately necessary to repeat the variable name in each stage :/

Dockerfile
fgreinacher: That's a common optimization to speed up Docker builds. It will cache the result of `npm install` and not execute it unless package.json or package-lock.json have changed.

README.md
fgreinacher: Yes, my main goal was also to not break "normal" usage 😀

I tested `BASE_URL` locally, yes!

Dockerfile
fgreinacher: Yes, sure! I added a comment!

fgreinacher:
> Also I wonder if we should use it here process.env.PUBLIC_URL to be more safe?

@magicmatatjahu I cannot comment inline, no idea why... Yes, that makes sense.

magicmatatjahu:
@fgreinacher Ok, I checked and everything works. Thanks very much for contribution! If you will have any problems with new docker image please ping me in this PR :)

magicmatatjahu (approved):

fgreinacher:
Works great now, thanks again for the quick review!

image

PR #271, opened 2022-02-21
fix: failing tests for newer versions of the converter

magicmatatjahu:
**Description**

- update `@asyncapi/converter` and `@asyncapi/react-component` to the latest version
- fix failing tests due to fact that `@asyncapi/converter` now handles the JSON spec 🚀 
- fix build of docker image, move `generate:template-parameters` script from `postinstall` to the `build` hook.

**Related issue(s)**
Fixes https://github.com/asyncapi/studio/pull/267
Fixes https://github.com/asyncapi/studio/pull/268
Fixes https://github.com/asyncapi/studio/pull/269
Fixes https://github.com/asyncapi/studio/pull/270

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,10 +26,10 @@
   ],
   "dependencies": {
     "@asyncapi/avro-schema-parser": "^1.0.1",
-    "@asyncapi/converter": "^0.7.0",
+    "@asyncapi/converter": "^0.9.0",
     "@asyncapi/openapi-schema-parser": "^2.0.1",
     "@asyncapi/parser": "^1.14.1",
-    "@asyncapi/react-component": "^1.0.0-next.33",
+    "@asyncapi/react-component": "^1.0.0-next.34",
     "@asyncapi/specs": "^2.13.0",
     "@headlessui/react": "1.4.1",
     "@hookstate/core": "^3.0.12",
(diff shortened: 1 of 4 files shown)


src/services/specification.service.ts
```diff
@@ -185,7 +184,7 @@ export class SpecificationService {
     );
   }
 
-  private static isNotSupportedVersion(rawSpec: string): boolean {
+  static isNotSupportedVersion(rawSpec: string): boolean {
```
boyney123, line 187: Is removing the `private` fixing the tests? From what I can understand it might not be? Any reason why that's changed? Unless I'm missing something 😅

package.json
```diff
@@ -50,7 +50,7 @@
   },
   "scripts": {
     "start": "cross-env DISABLE_ESLINT_PLUGIN=true craco start",
-    "build": "cross-env DISABLE_ESLINT_PLUGIN=true craco build",
+    "build": "npm run generate:template-parameters && cross-env DISABLE_ESLINT_PLUGIN=true craco build",
```
boyney123, line 53: Is this change for the docker build problem? How come it was moved from postinstall?

src/services/specification.service.ts
magicmatatjahu: That lines fix tests https://github.com/asyncapi/studio/pull/271/files#diff-30ced4fa50892c2cded728e0ed12cfdf2b0581861d73696ae02a56cdaf80c9a6R64-R68 :) I also checked that in one test we use `isNotSupportedVersion` function and TS complains that it's private so I changed it to public (only for tests), but it didn't cause any problems at all.

package.json
magicmatatjahu: Yeah, docker build process is broken (in master) because postinstall runs also in stage when we have copied only `package.json` and `package-lock.json` files -> https://github.com/asyncapi/studio/blob/master/Dockerfile#L9.

On `building` stage we have every needed deps/files to perform mentioned script :)

it was my mistake to add it in the `postinstall` hook.

boyney123:
Cool thanks for explaining @magicmatatjahu , from what understand it looks OK to me 👍

boyney123 (approved):

PR #280, opened 2022-02-28
chore: remove github-bot from codeowners

magicmatatjahu:
**Description**

Remove `github-bot` from codeowners.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -6,4 +6,4 @@
 
 # The default owners are automatically added as reviewers when you open a pull request unless different owners are specified in the file.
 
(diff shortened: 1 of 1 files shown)


derberg:
Context for those that will approve it:
- github after 3 years of Codeowners feature in production started validating the files
- current codeowners setup is broken, it is not possible to add user with `[]` in name 

We now remove the `github-actions[bot]`. And then we will test if even though we have branch protection that requires CODEOWNERS approval, `github-actions[bot]` approval will be accepted for merge too

mcturco (approved):

PR #282, opened 2022-03-01
feat: add a modal with information about what has changed in relation to the playground.

magicmatatjahu:
**Description**

Add a modal with information about what has changed in relation to the playground:
- modal opens when we have the query parameter `redirectedFrom=playground`

We need this to archive the playground repository as well as to make a redirect from playground to the studio - people should know what has changed in relation to the playground.

@derberg @alequetzalli Could you check the description of the changes?

Preview: https://deploy-preview-282--modest-rosalind-098b67.netlify.app/?redirectedFrom=playground

### DIFF
--- a/src/components/Modals/RedirectedModal.tsx
+++ b/src/components/Modals/RedirectedModal.tsx
@@ -50,7 +50,7 @@ export const RedirectedModal: React.FunctionComponent = () => {
 
   return (
     <ConfirmModal
-      title='Welcome in the AsyncAPI Studio!'
+      title='Welcome to the AsyncAPI Studio!'
       show={show}
       cancelText='OK'
       onCancel={onCancel}
--- a/src/components/Modals/RedirectedModal.tsx
+++ b/src/components/Modals/RedirectedModal.tsx
@@ -63,7 +63,7 @@ export const RedirectedModal: React.FunctionComponent = () => {
                   className='mx-auto rounded-md border border-transparent shadow-sm px-4 py-2 bg-pink-600 text-base font-medium text-white hover:bg-pink-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-pink-500 sm:text-sm'
                   onClick={onShowMoreClick}
                 >
-                  Show what's changed
+                  Show what&apos;s changed
                 </button>
               </div>
             </>
--- a/src/components/Modals/RedirectedModal.tsx
+++ b/src/components/Modals/RedirectedModal.tsx
@@ -36,6 +36,11 @@ export const RedirectedModal: React.FunctionComponent = () => {
   }, [show]); // eslint-disable-line
 
   function onCancel() {
+    if (typeof window.history.replaceState === 'function') {
+      const url = new URL(window.location.href);
+      url.searchParams.delete('redirectedFrom');
+      window.history.replaceState({}, window.location.href, url.toString());
+    }
     setShow(false);
   }
 
--- a/src/components/Modals/RedirectedModal.tsx
+++ b/src/components/Modals/RedirectedModal.tsx
@@ -1,9 +1,7 @@
 import React, { useEffect, useState } from 'react';
 
-// @ts-ignore
-import { Markdown } from '@asyncapi/react-component/lib/esm/components/Markdown';
-
 import { ConfirmModal } from './ConfirmModal';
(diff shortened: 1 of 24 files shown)


mcturco:
Agreed that we should redirect users from the playground to the studio and archive that repo. 

I am wondering if this modal can be improved in the future to serve as an onboarding screen that delivers the new features in smaller digestible steps rather than listing them all out in one screen (things like this can seem overwhelming to users, I wonder how many will even read down the whole list?) 

Also just wondering what prompted the addition of this modal? Did someone show some confusion between playground and studio? I'm interested to see if this new modal is useful to users coming from the playground or if all of these new features are a bit more self-explanatory. 

Also curious as to what the usage will look like of this query parameter... will it only be added when redirected from the playground domain or is this something that we would maybe add to a button where first-time users of studio are more likely to click on it? Curious about how we can get this onboarding information across to more users other than just redirecting from playground without it becoming a nuisance for returning users. 

(Sorry if that was off-topic from the goals of this PR. Just kind of thinking out loud about how we could improve this modal and use it to the Studio's advantage in other ways than just one)

magicmatatjahu:
@mcturco Thanks for that comment!

> I am wondering if this modal can be improved in the future to serve as an onboarding screen that delivers the new features in smaller digestible steps rather than listing them all out in one screen (things like this can seem overwhelming to users, I wonder how many will even read down the whole list?)

Yes, we can have the "changelog" like in (e.g.) Postman app to describe changes between versions of a studio. Could you create such issue? I don't know when we will implement it but it's worth keeping in mind.

> Also just wondering what prompted the addition of this modal? Did someone show some confusion between playground and studio? I'm interested to see if this new modal is useful to users coming from the playground or if all of these new features are a bit more self-explanatory.

Some people still use the playground and when we will enable redirection from playground to the studio, some people can be confused, so we need to list all changes in relation to the playground. So that list contains new features that we don't have in playground as well changes (that same features but working in different way),

> Also curious as to what the usage will look like of this query parameter... will it only be added when redirected from the playground domain or is this something that we would maybe add to a button where first-time users of studio are more likely to click on it? Curious about how we can get this onboarding information across to more users other than just redirecting from playground without it becoming a nuisance for returning users.

Only when people will be redirected. I wonder if we should save info that someone see that modal and don't show it again even if someone will redirected from playground again 🤔  For new users we can prepare such a tutorial what is something with highlighting relevant elements, but not in this PR.

magicmatatjahu:
@derberg 

> It gives anxiety seeing so much text at first 😂

I know as a society we're going back to cuneiform writing (emoji), but I had a problem with it myself. It can be cured with a psychologist  :trollface:

mcturco:
> Can we at least collapse this list of bullet points and offer a "Show what's changed" link? A simple one, something like a summary HTML tag, nothing fancy. It gives anxiety seeing so much text at first 😂

I have to agree with @fmvilas, I think this small change could improve the modal at least within this PR. But I will definitely create an issue for working on an onboarding process in Studio so we can work on that in the future 😄

magicmatatjahu:
@mcturco @fmvilas I pushed changes:

image

WDYT?

mcturco (approved):
@magicmatatjahu Good idea! Definitely less overwhelming 👍  LGTM 🎉

fmvilas:
Yeah, looks great to me. Just one more suggestion. Please change the button copy to "OK" instead of "Cancel" 🙏

src/components/Modals/RedirectedModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+    setShow(false);
+  }
+
+  function onShowMoreClick() {
+    setShowMore(true);
+  }
+
+  return (
+    <ConfirmModal
+      title='Welcome in the AsyncAPI Studio!'
```
fmvilas, line 48:
```suggestion
      title='Welcome to the AsyncAPI Studio!'
```

fmvilas (changes requested):
And last thing. Sorry for the cluttered review 😄

fmvilas (approved):
🚀 Ship it! :)

fmvilas:
@magicmatatjahu Just realized that after closing the modal, we leave the query param in the URL so if we refresh the page we get the modal again. It would be great to remove the query param (without refreshing) once we close the modal.

fmvilas (approved):

PR #342, opened 2022-05-20
chore: switch to visualiser library

### LINKED ISSUE #261: Split out visualizer
#### Reason/Context
I have raised this before but lost somewhere, or at least I cant find it 😅

I would like to have the Visualizer split out from Studio so we can start to use it and develop it independent of the studio.

#### Description
I already had the need for it and had to fork it from the studio and decouple it (can't recall what had to change, but something did). I especially needed it to show external applications from the point of view of the application.

billede

One of the questions that came up while doing this myself is when does Cupid come into the picture, and what should the visualizer do? 
billede

The distinction I did, for now, was that Visualizer rendered the view of a single application, and the external applications were a "subtle" link to the external application (see the first image). It can of course use Cupid to get this information about the relations to the application in view.

Cupid then showed the entire picture of the system and how everything was connected (see above image).

I know you said you did not want to split it just yet, but I would argue it should 😆

jonaslagoni:
**Description**
This PR switches to the split-out visualization library.

**Related issue(s)**
fixes #261


magicmatatjahu:
@jonaslagoni We probably should "wait" with this PR, due to simple reasons:
  
- styles atm aren't good, you can see:

  image

- we don't have "title" as we have in current implementation:

  image

- we should have also "flow" animation as we have in current implementation and we should add buttons for it (in top right corner):

  image

- we should have also additional buttons as we have in bottom left corner:

  image

I think that we should wait and merge it when we will move library to AsyncAPI org and fix above problems - of course PR can be still open :)

jonaslagoni:
> @jonaslagoni are you going to work on this? Do you need help?

If you could give me 48 hours a day, that would be really helpful 😆 

Current status if I am not mistaken:
- Still waiting for https://github.com/asyncapi/community/discussions/382 as we don't want to integrate it with Studio until then.
- Need to look into the styling and if anything is missing from the implementation.
- The buttons and titles that are missing are possible to add rather easily, so I just need to apply that.

Other then that, yes time is what I need 😆

jonaslagoni:
Alright, I fixed the issues you outlined @magicmatatjahu. I did not include any extra features as this is solely about switching internals with a library 🙂 

Any extra stuff I would suggest we add later.

jonaslagoni:
> we should have also "flow" animation as we have in current implementation and we should add buttons for it (in top right corner):

Okay I missed that one, that needs to be adapted.

magicmatatjahu:
@jonaslagoni Can we in some way fix styles for blocks. Look:

PR:

image

Master:

image

jonaslagoni:
I think those are valid improvements yes 👍

jonaslagoni:
This is what I am currently seeing in the PR. Do you see something else?

image

package.json
```diff
@@ -113,7 +114,7 @@
     "markdown-toc": "^1.2.0",
     "postcss": "^8.2.13",
     "raw-loader": "^4.0.2",
-    "react-scripts": "5.0.1",
+    "react-scripts": "4.0.3",
```
jonaslagoni, line 117: I changed the version because of #458

jonaslagoni:
@magicmatatjahu it's all ready now, switched the library to the official AsyncAPI library.

fmvilas (changes requested):
LGTM except this:

Captura de Pantalla 2022-11-20 a las 18 23 17


Please, make this text more readable :)

jonaslagoni:
Done @fmvilas 🙂

jonaslagoni:
Regarding the code smell it's because of different types that will clash because each library uses different parser versions. Choose to leave it as is, let me know if you want to disable the rule for that line. 

This is until the library support the new version: https://github.com/asyncapi/EDAVisualiser/issues/10

fmvilas (approved):
🚀

magicmatatjahu:
@jonaslagoni When I switch to the visualiser I see this in console and Studio are broken:

image

Which version of `react` and `react-flow` the EDAVisualiser uses? Studio now uses latest react (v18) and latest `react-flow` (v11) version.

Issue #252, opened 2022-01-31
studio.asyncapi.com only supports version 2.2.0


DJGummikuh:
#### Describe the bug
I am trying to edit an older async api yaml file that is based on Version 2.0.0 of the specification
this is failing because the editor only allows 2.2.0 to be used.

#### How to Reproduce
Open any file that is not based on the 2.2.0 specification
![image](https://user-images.githubusercontent.com/5188486/151785309-30db2c12-2307-4612-b454-df64b5ad153d.png)

#### Expected behavior
The editor should either support older standards, clearly define that it doesn't and (if that's the route to go) provide hints on what to do in these situations.

magicmatatjahu:
@DJGummikuh Hello! We have bug with that info about unsupported `2.0.0` version but it's still supported, trust me :) 

As I see in the "file preview" on the right side of the editor you have errors in your spec (check that red section at the bottom of your spec). You should click on the terminal section (at the bottom of the editor) to see errors and fix them:

![image](https://user-images.githubusercontent.com/20404945/151786666-67efcf3e-4f25-4bc3-939a-d72a8849cc1a.png)

derberg:
We should definitely not underline `2.0.0` with yellow saying `Value is not accepted` as it is misleading. It already prevented me from digging further for what are the other reasons to not render my document, cause if the version is not supported, why bother

fmvilas:
Also, leaving this comment from @DJGummikuh for the record: https://twitter.com/DJGummikuh/status/1488110968508928000. Looks like two different bugs? 🤔

magicmatatjahu:
@fmvilas If you have broken 2.0.0 spec you cannot make conversion. Converting 2.0.0 -> 2.2.0 is just a version change but still, it is not good to make conversion invalid spec and even if we will make it `2.2.0` will be still invalid.

DJGummikuh:
My issue is that the entire Document is simply red
I have no clue what exactly causes it and the marker at the bottom is also kinda unspecific :D
![image](https://user-images.githubusercontent.com/5188486/151790345-230026bb-a145-4d42-9a0d-8f9b50f9c4c5.png)

magicmatatjahu:
[...]

PR #352, opened 2022-06-01
fix: update monaco config even in invalid spec

magicmatatjahu:
**Description**

- update monaco config (version of spec) even in invalid spec.
- adjust config of monaco to read new json schemas for spec.
- update deps like `@asyncapi/converter` and templates.

**Related issue(s)**
Resolves https://github.com/asyncapi/studio/issues/252

### DIFF
--- a/src/react-app-env.d.ts
+++ b/src/react-app-env.d.ts
@@ -5,10 +5,8 @@ import { AsyncAPIDocument } from '@asyncapi/parser';
 
 declare global {
   interface Window {
-    // needed by monaco YAML plugin
+    // needed by monaco YAML plugin and Studio
     monaco: typeof monacoAPI;
-    // needed by Studio
-    Monaco: typeof monacoAPI;
     Editor: monacoAPI.editor.IStandaloneCodeEditor;
     MonacoEnvironment: monacoAPI.Environment | undefined;
     ParsedSpec: AsyncAPIDocument;
--- a/src/services/monaco.service.ts
+++ b/src/services/monaco.service.ts
@@ -61,7 +61,7 @@ export class MonacoService {
   }
 
   static loadLanguageConfig(asyncAPIVersion: string) {
-    const monacoInstance = window.Monaco;
+    const monacoInstance = window.monaco;
     if (!monacoInstance) return;
 
     const options = this.prepareLanguageConfig(asyncAPIVersion);
@@ -74,7 +74,7 @@ export class MonacoService {
   }
 
   static loadMonacoConfig() {
-    const monacoInstance = window.Monaco;
+    const monacoInstance = window.monaco;
     if (!monacoInstance) return;
 
     monacoInstance.editor.defineTheme('asyncapi-theme', {
@@ -100,7 +100,6 @@ export class MonacoService {
       loader.config({ monaco });
     }
     window.monaco = monaco;
-    window.Monaco = monaco;
 
     // load monaco config
     this.loadMonacoConfig();
--- a/src/services/editor.service.tsx
+++ b/src/services/editor.service.tsx
@@ -205,7 +205,7 @@ export class EditorService {
 
   static applyErrorMarkers(errors: any[] = []) {
     const editor = this.getInstance();
-    const Monaco = window.Monaco;
+    const Monaco = window.monaco;
 
     if (!editor || !Monaco) {
       return;


magicmatatjahu:
@mcturco Could you look at it?

magicmatatjahu:
@mcturco Yeah, you're right person for this, I mean for approval 😄 but if you want, you can make review of changes. I know that you probably don't understand the source code of that project but these changes are very needed.

package.json
```diff
@@ -31,7 +31,7 @@
     "@asyncapi/openapi-schema-parser": "^2.0.1",
     "@asyncapi/parser": "^1.15.1",
     "@asyncapi/react-component": "^1.0.0-next.38",
-    "@asyncapi/specs": "^3.0.0",
+    "@asyncapi/specs": "^2.14.0",
```
fmvilas, line 34: I don't think I understand why you're downgrading to a previous version of schemas. Can you elaborate on this?

src/services/specification.service.ts
```diff
@@ -45,6 +47,12 @@ export class SpecificationService {
         return asyncApiDoc;
       })
       .catch(err => {
+        try {
+          const asyncapiSpec = YAML.load(rawSpec);
+          MonacoService.updateLanguageConfig((asyncapiSpec as { asyncapi: string }).asyncapi);
```
fmvilas, line 52: Can't we do something like this?

```suggestion
          const asyncapiSpec: { asyncapi: string } = YAML.load(rawSpec);
          MonacoService.updateLanguageConfig(asyncapiSpec.asyncapi);
```

Or like this if that fails:

```suggestion
          const asyncapiSpec = YAML.load(rawSpec) as { asyncapi: string };
          MonacoService.updateLanguageConfig(asyncapiSpec.asyncapi);
```

fmvilas (changes requested):

package.json
magicmatatjahu: Doesn't anyone read the PR description? 😄 

> switch to old json schemas due to problem that monaco cannot fetch schemas from https://asyncapi.com/definitions...

However I try to make a workaround to support latest specs.

magicmatatjahu:
@fmvilas Could you check again?

package.json
fmvilas: I did read the PR description but I still don't understand the reason.

package.json
```diff
@@ -27,20 +27,20 @@
...
-    "@monaco-editor/react": "^4.2.2",
+    "@monaco-editor/react": "^4.4.5",
     "@semantic-release-plus/docker": "^3.1.1",
     "js-base64": "^3.7.2",
     "js-file-download": "^0.4.12",
     "js-yaml": "^4.1.0",
-    "monaco-editor": "^0.28.1",
-    "monaco-yaml": "^2.5.1",
+    "monaco-editor": "0.28.1",
+    "monaco-yaml": "2.5.1",
```
fmvilas, line 43: What is the reasoning behind fixing these versions?

package.json
magicmatatjahu: Look on latest changes.

src/react-app-env.d.ts
```diff
@@ -5,6 +5,9 @@ import { AsyncAPIDocument } from '@asyncapi/parser';
 
 declare global {
   interface Window {
+    // needed by monaco YAML plugin
+    monaco: typeof monacoAPI;
```
fmvilas, line 9: `monaco` and `Monaco`? Can't we make it just one? Maybe make Studio get `monaco` instead?

src/services/monaco.service.ts
```diff
@@ -89,8 +89,18 @@ export class MonacoService {
...
+    let monaco: typeof monacoAPI;
+
+    // JEST cannot bundle monaco-editor in test environment so we need to fetch that package from cdn
+    // in dev or production environment we will use bundled monaco-editor
+    if (process.env.NODE_ENV === 'test') {
+      monaco = await loader.init();
+    } else {
+      monaco = await import('monaco-editor');
+      loader.config({ monaco });
+    }
```
fmvilas, line 101: I don't have much idea about the Monaco internals but every time I see code that needs to be changed to make tests work, it gives me shivers 😄 Isn't there really a way to make tests work without having to make this code aware of the tests?

magicmatatjahu: I tried to map the imports to the appropriate modules, but this didn't work at all. Jest is building the project before testing (I was surprised) making monaco-editor mapping difficult. You can look for issues on github or even on stackoverflow where a lot of people have problems with this package on the jest side. I know that such ifs should not be, but this is the simplest solution to the problem.

package.json
magicmatatjahu: Higher versions of `monaco-editor` have breaking changes, which are not compatible with given version of `monaco-yaml`. The latest version of `monaco-yaml` supports newer versions of monaco, but here again there are problems related to the worker on the yaml side. Each language in monaco is divided into 2 parts: the code on the main thread side, and the worker, which executes code not on the main thread, but on a separate thread to optimize the process (linting etc). Newer versions of `monaco-yaml` creates this worker differently, and to use it I would have to eject react-scripts and change the webpack. I want to fix this problem but not in this PR. Currently on the hosted studio `monaco-yaml` is not working fully (worker doesn't work) and I would prefer not to wait with these supports and just publish the fixed studio and then deal with dependencies.

fmvilas: That makes sense. Thanks or the explanation 👍

fmvilas (approved):
LGTM 👍 Thanks for the explanations!

PR #359, opened 2022-06-07
feat: add survey popover

### LINKED ISSUE #358: Add CTA for user survey that we are currently running
We are currently running a survey as mentioned in this issue: https://github.com/asyncapi/asyncapi-react/issues/593

I am proposing that we add a call to action within the UI of the Studio to attract more survey participants. I would like to work with the other code maintainers of this repo to see what components we already have in the UI that we might be able to use for this CTA.

I'm thinking that we just need the following elements:
- Main title of CTA (Something like: "Help us improve AsyncAPI Studio"
- Short body copy: We know that the best way to make our tools better is to understand our users better. Help us define your needs by completing this short survey!
- Button with link to survey
- Bonus: maybe I can come up with a simple illustration/graphic that will serve as a visual incentive for someone to pay attention to what we are asking in the CTA

Open to suggestions on this as well, but would like to get this up ASAP!

cc: @magicmatatjahu

magicmatatjahu:
**Description**

- Add popover with survey info in bottom left corner
- That popover disappears after first close/go to survey - if you wanna see it again, clear `localStorage` cache from given site.
- as Studio works only on desktop, I didn't implement the mobile styles

@mcturco Thanks for mocks! :)

**Related issue(s)**
Resolves #358


mcturco:
Thanks @magicmatatjahu for producing this so quick! Just a couple things before we merge:

- I think we should add maybe a 3 second timeout delay and then fade in the survey popover
- let's change the hover class of the button to `bg-gray-50` to match the buttons that are similar in the other modals in Studio

After those couple of changes I think this is perfect and we should merge! 😄

mcturco (approved):
LGTM 🚢

Issue #403, opened 2022-08-21
load spec file via Docker volume


steevehook:
#### Reason/Context

I want to be able to spin up a docker container that has the `asyncapi.yaml` file either passed as a bind mount or env variable, to be able to load that as soon as the container starts, instead of doing that step manually

#### Description

As a user of this Docker image I wanna be able to reference a Docker volume or env variable and automatically load the configuration on container start.
If this is already present, at least document this feature, I had to clone the project, look it all up and still could not find a proper solution for this

magicmatatjahu:
@steevehook Hi! Thanks for the issue! Atm we don't have such a feature for Docker image. We are open for contributions, so if you wanna contribute, you're welcome :)

steevehook:
Then I guess, PR incoming 😄

steevehook:
That one time when you wanna contribute to Open Source and you got the wrong laptop, or at least that what it says when I try to run `npm run generate:docker`
```
#11 95.83 npm ERR! code 1
#11 95.83 npm ERR! path /node_modules/puppeteer
#11 95.83 npm ERR! command failed
#11 95.83 npm ERR! command sh -c node install.js
#11 95.83 npm ERR! The chromium binary is not available for arm64.
#11 95.83 npm ERR! If you are on Ubuntu, you can install with: 
#11 95.83 npm ERR! 
#11 95.83 npm ERR!  sudo apt install chromium
#11 95.83 npm ERR! 
#11 95.83 npm ERR! 
#11 95.83 npm ERR!  sudo apt install chromium-browser
#11 95.83 npm ERR! 
#11 95.83 npm ERR! ERROR: Failed to set up Chromium r1002410! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
#11 95.83 npm ERR! Error
#11 95.83 npm ERR!     at handleArm64 (/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:141:11)
#11 95.83 npm ERR!     at BrowserFetcher.download (/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:286:13)
#11 95.83 npm ERR!     at async downloadBrowser (/node_modules/puppeteer/lib/cjs/puppeteer/node/install.js:56:5)
#11 95.87 
#11 95.87 npm ERR! A complete log of this run can be found in:
[...]

PR #404, opened 2022-08-22
feat: import spec from file on editor init

steevehook:
**Description**

The purpose of this PR is to allow the users of the Studio Docker image to be able to initialise the container with the spec of their choice by providing a Docker bind mount.

**Related issue(s)**

Resolves [#403 ](https://github.com/asyncapi/studio/issues/403#issue-1345519106)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,12 @@ Run:
 
 ```bash
 docker run -it -p 8000:80 asyncapi/studio
+docker run -it -p 8000:80 -v "$(pwd)/another/path/to/asyncapi:/usr/share/nginx/html/docs/asyncapi" asyncapi/studio
 ```
 
 and then go to [http://localhost:8000](http://localhost:8000).
+
+Also feel free to check out the [docker-compose.yml](./docker-compose.yml) example.
 
 The `asyncapi/studio` image is based on the official `nginx` image.
 Please refer to the [Nginx documentation](https://registry.hub.docker.com/_/nginx/) to learn how to e.g. pass a custom `nginx` configuration or plug in additional volumes.
--- a/src/components/Editor/MonacoWrapper.tsx
+++ b/src/components/Editor/MonacoWrapper.tsx
@@ -28,6 +28,7 @@ export const MonacoWrapper: React.FunctionComponent<MonacoWrapperProps> = ({
     // save editor instance to the window
     window.Editor = editor;
     // parse on first run the spec
+    await EditorService.importFromURL('docs/asyncapi/asyncapi.yaml');
     SpecificationService.parseSpec(EditorService.getValue());
 
     // apply save command
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ ARG BASE_URL_PLACEHOLDER=189b303e-37a0-4f6f-8c0a-50333bc3c36e
 FROM docker.io/library/node:16.13.2 as build
 
 ARG BASE_URL_PLACEHOLDER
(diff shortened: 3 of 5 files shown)


README.md
```diff
@@ -49,10 +49,13 @@ Run:
 
 ```bash
 docker run -it -p 8000:80 asyncapi/studio
+docker run -it -p 8000:80 -v "$(pwd)/another/path/to/asyncapi:/usr/share/nginx/html/docs/asyncapi" asyncapi/studio
```
magicmatatjahu, line 52: Could we split that line to separate paragraph to describe how to add volume with AsyncAPI spec, something like:

```md
To mount a specification, use the command:

\```bash
docker run -it -p 8000:80 -v "$(pwd)/another/path/to/asyncapi:/usr/share/nginx/html/docs/asyncapi" asyncapi/studio
\```

> **NOTE**: The file should be named `asyncapi.yaml` or `asyncapi.json` in the specified location.
```

Please remove `\` before backticks 😅

README.md
```diff
@@ -49,10 +49,13 @@ Run:
...
 ```
 
 and then go to [http://localhost:8000](http://localhost:8000).
 
+Also feel free to check out the [docker-compose.yml](./docker-compose.yml) example.
```
magicmatatjahu, line 57: Could we paste the value of this file in the readme under the section with docker and describe how to use the studio with docker-compose? Until now, we have not had any use case that we need examples and for now I do not see the point of having them (I mean `examples` folder or something similar).

Something like:

```md
Example of using `docker-compose`:

/```yaml
services:
  async-studio:
    build: .
    container_name: events-api
    restart: on-failure
    ports:
      - "8000:80"
    volumes:
      - ./public/docs/asyncapi:/usr/share/nginx/html/docs/asyncapi # when you need to mount specification
/```
```

WDYT?

src/components/Editor/MonacoWrapper.tsx
```diff
@@ -28,6 +28,7 @@ export const MonacoWrapper: React.FunctionComponent<MonacoWrapperProps> = ({
     // save editor instance to the window
     window.Editor = editor;
     // parse on first run the spec
+    await EditorService.importFromURL('docs/asyncapi/asyncapi.yaml');
```
magicmatatjahu, line 31: There I have a big problem. You try to parse given mounted spec every time, so if someone will have spec saved in the localStorage, we will always override it. Due to the fact that you provided `asyncapi.yaml` in this PR, you can go to the https://deploy-preview-404--modest-rosalind-098b67.netlify.app/ preview, change something in the editor, then reload, and you will see that your changes disappeared - it's a breaking change of current solution. You can also check www.studio.asyncapi.com how now it works :)

My proposition:
- remove provided `asyncapi.yaml` in this PR
- check if given spec exist or not, by try catch - you probably need to change some logic inside `EditorService.importFromURL` function, of course I can help you with that. Let me know.
- move that line of code to the https://github.com/asyncapi/studio/blob/master/src/services/navigation.service.ts#L83 I can help with that because there are some edge cases related to the `url` and `base64` query parameters and "readOnly" mode of studio.

WDYT?

magicmatatjahu (changes requested):
Awesome! However I described one big problem in your solution. Also we need to a bit to improve the Readme :) Let me know about any problems.

README.md
steevehook: idk who or what added that back slash 🤣 will immediately cover the shame

magicmatatjahu:
@steevehook Hi! Do you need help?

PR #408, opened 2022-09-05
build: support arm64 architecture in docker image.

### LINKED ISSUE #407: Support arm64 architecture in docker image
#### Reason/Context

Currently the docker image fails to start in a machine with `arm64` architecture (e.g. Macs with ARM processors). Even though docker supports virtualization, the `nginx` for `amd64` architectures [fails to start](https://stackoverflow.com/questions/68704538/error-io-setup-failed-38-function-not-implemented-in-nginx-for-arm-m1):

```
docker run -it -p 8000:80 asyncapi/studio                        
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
/docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
/docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
/docker-entrypoint.sh: Launching /docker-entrypoint.d/set-public-url.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
10-listen-on-ipv6-by-default.sh: info: Getting the checksum of /etc/nginx/conf.d/default.conf
10-listen-on-ipv6-by-default.sh: info: Enabled listen on IPv6 in /etc/nginx/conf.d/default.conf
/docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
/docker-entrypoint.sh: Launching /docker-entrypoint.d/30-tune-worker-processes.sh
/docker-entrypoint.sh: Configuration complete; ready for start up
2022/09/05 21:18:43 [notice] 1#1: using the "epoll" event method
2022/09/05 21:18:43 [notice] 1#1: nginx/1.21.5
2022/09/05 21:18:43 [notice] 1#1: built by gcc 10.3.1 20211027 (Alpine 10.3.1_git20211027) 
2022/09/05 21:18:43 [notice] 1#1: OS: Linux 5.10.104-linuxkit
2022/09/05 21:18:43 [notice] 1#1: getrlimit(RLIMIT_NOFILE): 1048576:1048576
2022/09/05 21:18:43 [notice] 1#1: start worker processes
2022/09/05 21:18:43 [notice] 1#1: start worker process 89
2022/09/05 21:18:43 [notice] 1#1: start worker process 91
2022/09/05 21:18:43 [notice] 1#1: start worker process 93
2022/09/05 21:18:43 [notice] 1#1: start worker process 94
2022/09/05 21:18:43 [notice] 1#1: start worker process 96
2022/09/05 21:18:43 [notice] 1#1: start worker process 97
2022/09/05 21:18:43 [emerg] 89#89: io_setup() failed (38: Function not implemented)
2022/09/05 21:18:43 [emerg] 91#91: io_setup() failed (38: Function not implemented)
2022/09/05 21:18:43 [emerg] 94#94: io_setup() failed (38: Function not implemented)
2022/09/05 21:18:43 [emerg] 97#97: io_setup() failed (38: Function not implemented)
2022/09/05 21:18:43 [emerg] 93#93: io_setup() failed (38: Function not implemented)
2022/09/05 21:18:43 [emerg] 96#96: io_setup() failed (38: Function not implemented)
```

#### Description

In order to support the `arm64` architecture, no code changes are required. Only changes to the build automation so that the docker image is also built for that architecture, since the base images used in the `Dockerfile` already have builds for `arm64` widely available.

I will submit a pull request with proposed changes.

sandoval:
**Description**

Support arm64 architecture in docker image.

- Chromium was skipped from puppeteer since it was already unused and it isn't compatible with `arm64`. Chromium would only be used to export a PDF of the HTML site, which is already unsupported. If support is required in the future, install chromium using the package manager of the image.
- Changes were tested in a Mac with M1 processor and worked successfully with local docker build.
- Changes to the github workflows are untested due to missing test environment.

Resolves #407

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -80,6 +80,12 @@ jobs:
       - if: steps.packagejson.outputs.exists == 'true'
         name: Install dependencies
         run: npm install
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: 'linux/amd64,linux/arm64'
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
       - if: steps.packagejson.outputs.exists == 'true'
         name: Publish to any of NPM, Github, and Docker Hub
         id: release
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,9 @@ FROM docker.io/library/node:16.13.2 as build
 
 ARG BASE_URL_PLACEHOLDER
 
+# Skip puppeteer download since chromium isn't used by the project.
+ARG PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1
+
 COPY package.json package-lock.json ./
 RUN npm install --frozen-lockfile
 
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "generate:readme:toc": "markdown-toc -i README.md",
     "generate:assets": "npm run build && npm run generate:readme:toc",
     "generate:template-parameters": "ts-node ./scripts/template-parameters.ts",
-    "generate:docker": "docker build --tag asyncapi/studio .",
+    "generate:docker": "docker buildx build --platform linux/amd64,linux/arm64 --tag asyncapi/studio .",
     "bump:version": "npm --no-git-tag-version --allow-same-version version $VERSION",
     "release": "npm run generate:docker && semantic-release",
     "prepublishOnly": "npm run build && npm run generate:readme:toc"
--- a/.github/workflows/if-nodejs-pr-testing.yml
+++ b/.github/workflows/if-nodejs-pr-testing.yml
@@ -33,6 +33,12 @@ jobs:
         id: packagejson
         run: test -e ./package.json && echo "::set-output name=exists::true" || echo "::set-output name=exists::false"
         shell: bash
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: 'linux/amd64,linux/arm64'
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
       - if: steps.packagejson.outputs.exists == 'true'
         name: Setup Node.js
         uses: actions/setup-node@v2


Dockerfile
```diff
@@ -5,6 +5,9 @@ FROM docker.io/library/node:16.13.2 as build
 
 ARG BASE_URL_PLACEHOLDER
 
+# Skip puppeteer download since chromium isn't used by the project.
+ARG PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1
```
magicmatatjahu, line 9: Thanks for that! For info: one of dev-dependency needs that, but we only use from that package only some infos from its `package.json`, so `puppeteer` is needed.

.github/workflows/if-nodejs-release.yml
```diff
@@ -80,6 +80,12 @@ jobs:
       - if: steps.packagejson.outputs.exists == 'true'
         name: Install dependencies
         run: npm install
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: 'linux/amd64,linux/arm64'
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
```
magicmatatjahu, line 88: I cannot accept such a changes. Almost all workflows in the `.github/workflows` folder are copied (by our one global workflow) from this place - https://github.com/asyncapi/.github/tree/master/.github/workflows, so it means that if someone will update workflows in `.github` repo, updated workflows in this PR will be overwritten.

If you need to add such a setup in workflow, you should add these steps in appropriate workflows in the https://github.com/asyncapi/.github/tree/master/.github/workflows. Update `generate:docker` script is ok, because every repo can perform different logic for assets generation.

cc @derberg What do you think? Should we move these steps to the `.github` repo? I ask because I know that you created PR with some updates on that workflows related to the automatic release and also that PR made changes related to docker release. How you think, how we should do this? @sandoval Could you wait for @derberg answer?

package.json
```diff
@@ -59,7 +59,7 @@
     "generate:readme:toc": "markdown-toc -i README.md",
     "generate:assets": "npm run build && npm run generate:readme:toc",
     "generate:template-parameters": "ts-node ./scripts/template-parameters.ts",
-    "generate:docker": "docker build --tag asyncapi/studio .",
+    "generate:docker": "docker buildx build --platform linux/amd64,linux/arm64 --tag asyncapi/studio .",
```
magicmatatjahu, line 62: I am not an expert of docker, so could you describe me this change? I mean, I have problem to understand how many (is it a good question) docker images will be created, with different architecture destination, or everything will be virtualised across one image? Thanks for explanation! I am afraid about backward compatibility.

magicmatatjahu (commented):
@sandoval Thanks for PR! I added some comments that we should discuss first :)

.github/workflows/if-nodejs-release.yml
derberg: @sandoval as @magicmatatjahu wrote, whatever you change here, will anyway be overwritten (have a look at the comment at the beginning of the workflow.

yeah, so at the moment, I'm rebuilding a bit the release pipeline to have it less error-prone. The idea is to handle docker release after main release, like we do it here https://github.com/asyncapi/github-action-for-generator/blob/master/.github/workflows/release-docker.yml (example ob publishing docker after GitHub release)

so @magicmatatjahu @sandoval, this is what I suggest you do, really. Have a separate workflow that handles docker release.  Design it as generic as possible, call file `if-docker-release.yml`. Then once we add it in the `.github` repo, we will just override and long-term maintain from there. I hope is makes sense what I'm writing 🤔

magicmatatjahu: So as I understand @derberg, that new workflow should be created in this PR, and then we should move it to the `.github` repo, right? Or should we start from creating workflow in the `.github` repo.

derberg: I definitely suggest doing it here first and having it battle-tested here, we can move it over after a release or 2 to `.github` repo alone, or @sandoval can do it if he is interested

sandoval: Thank you for the comments @magicmatatjahu and @derberg ! I also think it would be best to have a separate workflow for building the docker image, based on the github actions for building such images. However, I tried to adapt to the project's logic when submitting the PR.
In using the github actions, I would use [docker-build-push](https://github.com/docker/build-push-action) action instead of `semantic-release`, what do you think?

package.json
sandoval: Of course! I'll try my best to explain.

"Docker Buildx is a CLI plugin that extends the docker command with the full support of the features provided by [Moby BuildKit](https://github.com/moby/buildkit) builder toolkit." [[1](https://docs.docker.com/build/buildx/)].

This is the way docker enables us to cross-compile docker images for different architectures. By using the `docker-container` Buildkit driver, the builds run inside a buildkit container, which may have support for QEMU emulation [[2](https://en.wikipedia.org/wiki/QEMU)] in order to compile for other architectures.

Other change that is a side effect is using docker Buildkit [[3](https://docs.docker.com/develop/develop-images/build_enhancements/)] as the build system, which was introduced in the 18.09 version and is currently the recommended build system to use. It also enables us to use new features in the Dockerfile, such as caching.

.github/workflows/if-nodejs-release.yml
magicmatatjahu: @sandoval It will be problematic to run in two independent jobs `semantic-release`, so yeah, `[docker-build-push](https://github.com/docker/build-push-action)` should be better. However, next to `latest` version we should also push version indicated by `package.json`.

package.json
magicmatatjahu: @sandoval Ok, thank you very much! So tl;dr; it is a emulation 😄

.github/workflows/if-nodejs-release.yml
derberg: yup, forget semantic-release

as you will react to release event, you will not have access yet to updated package.json with proper version. just look at the action I linked few comments earlier. there is an example how I get package version from release tag, just copy it

PR #426, opened 2022-09-19
feat: add arg DOCUMENT_URL in Dockerfile

rikuson:
**Description**

This is continued from https://github.com/asyncapi/studio/pull/404 .
I caught up another solution: passing enviroment variable from `docker-compose` .
Please take a look.

**Related issue(s)**

Resolves https://github.com/asyncapi/studio/issues/403


rikuson:
Hmm, I think it doesn't work.
I just misunderstood.

docker-compose can't build without local Dockerfile.

PR #434, opened 2022-09-27
feat: integrate new ParserJS with Spectral

magicmatatjahu:
**Description**

- Integrate new v2 ParserJS with Spectral support

TODO: write unit tests and expose from ParserJS the `getRangeForJSOnPointer` functionality.

Example:

image

**Related issue(s)**
Part of https://github.com/asyncapi/parser-js/issues/481


smoya:
Netlify build fails because of https://github.com/asyncapi/parser-js/pull/650

smoya:
@magicmatatjahu can you please bump `@asyncapi/parser` to `2.0.0-next-major.6` so we fix Netlify build? (I have no permissions to push to your branch unfortunately).

jonaslagoni (changes requested):
Love the new error handling! 🔥 And how accurately it highlights the areas that has problems.

Hate the warning system though 😆 As you are entering the realm of governance rules, which highly vary from user to user. I would HATE to see warnings for an arbitrary rule that someone says I should support, for example, `Info object must have "contact" object.`, as it's not a requirement from the AsyncAPI spec, but spectral. 

For some, this will become highly confusing.

I suggest anything beyond requirements from the spec should be turned off, until (if it's being planned) it's possible to personally select linting rules.

And it makes the editor almost unusable with yellow markings all over 😄 Which almost forces you to take action on the warnings.
image

Also, are warnings supposed to be defined as problems? 🤔 
image

magicmatatjahu:
@jonaslagoni Thanks for quick review!

> Hate the warning system though 😆 As you are entering the realm of governance rules, which highly vary from user to user. I would HATE to see warnings for an arbitrary rule that someone says I should support, for example, Info object must have "contact" object., as it's not a requirement from the AsyncAPI spec, but spectral.

Warnings inform you about some good practices and we should enable them by default - it's not treated as error and studio will render html documentation on the right, it's only a info to user that something is a good practice 😄 Of course description of such a warning should be defined as `Info object should have "contact" object.` not "must" but it's a "bug" on Spectral side not our. I don't wanna wait for fixing such a thing on Spectral side, because we will make integration next 1-2 months and it's not a big bug at all.

> I suggest anything beyond requirements from the spec should be turned off, until (if it's being planned) it's possible to personally select linting rules.

Yeah, custom rules and overrides are in plan, but probably it will be enabled in next year.

> And it makes the editor almost unusable with yellow markings all over 😄 Which almost forces you to take action on the warnings.

Yeah, I hat it either 😆 so I added suggestion (if you didn't see that) in slack:

> The only problem I have with the current integration is that we underline even warnings/hints in the editor and not just errors (which should be underlined), making the document itself in the editor have too many "colors" and difficult to read, even document may be correct. Should we inform about warnings and hints in the left side of the editor (where the line numbers are) with warning and hint icons? What do you think about this.

So we should underline only errors (by red color) and warning and hints should be displayed in the left panel (where lines number are) as icon - example: https://microsoft.github.io/monaco-editor/playground.html#interacting-with-the-editor-rendering-glyphs-in-the-margin check this red square in editor.

---

> Also, are warnings supposed to be defined as problems? 🤔

We use monaco-editor, this same which VSC uses, so "warnings", "hints" etc are also considering as "problems" - VSC treat every error/waring/hint as problem not "diagnostic" as we do and I don't think so that we can change description of such a warnings.

jonaslagoni:
> Warnings inform you about some good practices and we should enable them by default

I understand that it makes sense to have good practices highlighted, but the matter of the facts is that its only relevant for some users.

If you develop an internal application and use AsyncAPI to describe it and you are a small team, adding contact info makes no sense.

^just an example, sure in some cases it might be, but for small teams or solos, it does not and does nothing but create friction 🙂 

If you really want it, I suggest it be added as part of the settings so it can be turned off at the will of the user.

> Of course description of such a warning should be defined as Info object should have "contact" object. not "must" but it's a "bug" on Spectral side not our. I don't wanna wait for fixing such a thing on Spectral side, because we will make integration next 1-2 months and it's not a big bug at all.

That's understandable 🙂 

> So we should underline only errors (by red color) and warning and hints should be displayed in the left panel (where lines number are) as icon - example: https://microsoft.github.io/monaco-editor/playground.html#interacting-with-the-editor-rendering-glyphs-in-the-margin check this red square in editor.

Hard to say 😆 Cause if you then have to add one for each line where the yellow is currently highlighted, it's also quite a lot still 😄

magicmatatjahu:
> I understand that it makes sense to have good practices highlighted, but the matter of the facts is that its only relevant for some users. And it comes down to their use-case and what they would like to see as good practice.

> If you really want it, I suggest it be added as part of the settings so it can be turned off at the will of the user.

I understand but I don't like such "filtering", because warning contact is indeed unnecessary but for `operationId` that operation should define is good imho 😄 I personally would leave it, for me it's no problem, but I can add options to enable disabling these warnings. I will make it, but probably not today 😄 

> Hard to say 😆 Cause if you then have to add one for each line where the yellow is currently highlighted, it's also quite a lot still 😄

That icon will be only visible in starting line, so if your object is defined in 20 lines you will only see that icon on first line.

jonaslagoni:
> I understand but I don't like such "filtering", because warning contact is indeed unnecessary but for `operationId` that should define is good imho 😄 I personally would leave it, for me it's no problem, but I can add options to enable/disable these warnings. I will make it, but probably not today 😄

Just a global one for all warnings I think should be enough, for example, `disable good practice warnings` or something like that.

> That icon will be only visible in starting line, so if your object is defined in 20 lines you will only see that icon on first line.

That would probably make sense then 👍

magicmatatjahu:
I added vertical lines in the left side of editor (where lines are):

image

Errors are still visible as markers:

image 

cc @smoya @jonaslagoni what do you think?

src/components/Terminal/ProblemsTab.tsx
```diff
@@ -1,73 +1,275 @@
...
+            <button type='button' className={`hover:bg-gray-900 rounded-sm border border-transparent ${search ? 'opacity-100' : 'opacity-0'}`} onClick={() => {
+              if (inputRef.current) {
+                inputRef.current.value = '';
+              }
+              setSearch('');
+            }}>
+              <VscClose />
+            </button>
+          </div>
+          <a href='https://stoplight.io/open-source/spectral' title='Spectral website' target='_blank' rel="noreferrer" className='text-white hover:text-blue-500'>
```
derberg, line 219:
```suggestion
          <a href='https://github.com/stoplightio/spectral' title='Spectral Github repo' target='_blank' rel="noreferrer" className='text-white hover:text-blue-500'>
```

magicmatatjahu:
@derberg Ok I will do that, thanks for suggestion!

src/components/Modals/Settings/SettingsModal.tsx
```diff
@@ -110,6 +149,71 @@ export const SettingsModal: React.FunctionComponent = () => {
...
+      content: (
+        <>
+          <div>
+            <div className="flex flex-col mt-4 text-sm">
+              <div className="flex flex-row content-center justify-between">
+                <label
+                  htmlFor="settings-governance-show-warnings"
+                  className="flex justify-right items-center w-1/2 content-center font-medium text-gray-700"
+                >
+                  Show&nbsp;<strong>warning</strong>&nbsp;governance issues:
```
fmvilas, line 164:
```suggestion
                  Show&nbsp;<strong>warning</strong>&nbsp;governance issues
```

src/components/Modals/Settings/SettingsModal.tsx
```diff
@@ -110,6 +149,71 @@ export const SettingsModal: React.FunctionComponent = () => {
...
+            </div>
+          </div>
+          <div>
+            <div className="flex flex-col mt-4 text-sm">
+              <div className="flex flex-row content-center justify-between">
+                <label
+                  htmlFor="settings-governance-show-informations"
+                  className="flex justify-right items-center w-1/2 content-center font-medium text-gray-700"
+                >
+                  Show&nbsp;<strong>information</strong>&nbsp;governance issues:
```
fmvilas, line 183:
```suggestion
                  Show&nbsp;<strong>information</strong>&nbsp;governance issues
```

src/components/Modals/Settings/SettingsModal.tsx
```diff
@@ -110,6 +149,71 @@ export const SettingsModal: React.FunctionComponent = () => {
...
+                <Switch
+                  toggle={governanceWarnings}
+                  onChange={setGovernanceWarnings}
+                />
+              </div>
+              <div className='text-gray-400 text-xs'>
+                Show warning governance issues in the editor&apos;s&nbsp;<strong>Diagnostics</strong>&nbsp;tab.
+              </div>
+            </div>
+          </div>
```
fmvilas, line 175: I think you can already convert this into a component. It's repeated 3 times.

src/components/Terminal/ProblemsTab.tsx
```diff
@@ -1,73 +1,288 @@
...
+  );
+  case 3: return (
+    <div className='flex flex-row items-center justify-center'>
+      <VscLightbulb className='text-green-500 w-4 h-4' />
+    </div>
+  );
+  default: return (
+    <div className='flex flex-row items-center justify-center'>
+      <VscError className='text-red-500 w-4 h-4' />
     </div>
```
fmvilas, line 93: It would be awesome to have tooltips in these icons. Until I came to the code, I couldn't figure out what the green light bulb meant.

magicmatatjahu: I hope that you like it :)

image

src/components/Modals/Settings/SettingsModal.tsx
magicmatatjahu: Done :)

src/components/Modals/Settings/SettingsModal.tsx
```diff
@@ -110,6 +149,71 @@ export const SettingsModal: React.FunctionComponent = () => {
...
+          <div>
+            <div className="flex flex-col mt-4 text-sm">
+              <div className="flex flex-row content-center justify-between">
+                <label
+                  htmlFor="settings-governance-show-hints"
+                  className="flex justify-right items-center w-1/2 content-center font-medium text-gray-700"
+                >
+                  Show&nbsp;<strong>hint</strong>&nbsp;governance issues:
```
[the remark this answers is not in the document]
magicmatatjahu: Done :)

src/components/Modals/Settings/SettingsModal.tsx
magicmatatjahu: Done :)

src/components/Modals/Settings/SettingsModal.tsx
magicmatatjahu: Done :)

magicmatatjahu:
@fmvilas I applied all suggestions. Could you check again? Thanks!
@derberg You added suggestion to have button in tab to open the `Governance settings`. Could you check and let me know if you like it. I'll be very grateful :)

derberg:
works like a charm ❤️

fmvilas (approved):
🐑 🇮🇹

magicmatatjahu:
@fmvilas Conflicts are resolved. Could you accept again?

src/components/Terminal/ProblemsTab.tsx
```diff
@@ -1,73 +1,299 @@
...
+  const activeBg = 'bg-gray-900';
+  const notActiveBg = 'bg-gray-700';
+
+  const setActive = useCallback((type: ActiveSeverity) => {
+    defaultSetActive(type === active ? 'all' : type);
+  }, [active]);
+
+  return (
+    <ul className='flex flex-row items-center'>
+      <Tooltip content="Show only errors">
```
fmvilas, line 124: The tooltip is great if this option is not selected but when I click on the button it still says "Show only errors", and it's wrong. It should probably say "Show everything" or "Stop showing only errors" or something like that.

That said, playing with these buttons I realized it's acting as a "select" or "option". Wouldn't it make more sense to accept selecting multiple options? For instance, I want to see errors and warnings only. This is a common thing when you're fixing problems and don't want to have much noise. Also, tooltips will become easier. They'd be something like "Show errors" or "Hide errors".

src/components/Terminal/ProblemsTab.tsx
```diff
@@ -1,73 +1,299 @@
...
+
+  const activeBg = 'bg-gray-900';
+  const notActiveBg = 'bg-gray-700';
+
+  const setActive = useCallback((type: ActiveSeverity) => {
+    defaultSetActive(type === active ? 'all' : type);
+  }, [active]);
+
+  return (
+    <ul className='flex flex-row items-center'>
```
fmvilas, line 123: Why are you using `<ul>` if then you're not using `<li>` inside? 🤔

src/components/Terminal/ProblemsTab.tsx
```diff
@@ -1,73 +1,299 @@
...
+          className={`disabled:cursor-not-allowed w-full inline-flex justify-center border border-transparent shadow-xs px-2 py-1 ml-px ${active === 1 ? activeBg : notActiveBg} text-xs font-medium text-white hover:bg-gray-900 disabled:bg-gray-700 focus:outline-none focus:ring-1 focus:ring-offset-1 focus:ring-gray-700`}
+          onClick={() => setActive(DiagnosticSeverity.Warning)}
+        >
+          <div className='flex flex-row items-center justify-center'>
+            <SeverityIcon severity={1} />
+            <span className='ml-1'>{warningDiagnostics.length}</span>
+          </div>
+        </button>
+      </Tooltip>
+      <Tooltip content="Show only informations">
```
fmvilas, line 148:
```suggestion
      <Tooltip content="Show only information messages">
```

fmvilas (changes requested):
Getting better but still got UX issues.

src/components/Terminal/ProblemsTab.tsx
magicmatatjahu: I don't know, probably I missed, fixed 😅

src/components/Terminal/ProblemsTab.tsx
magicmatatjahu: Ok, I implemented this second approach with `Show all diagnostics` text if user select only one option :) Play with it and you will see what I have in mind. Thanks for suggestion!

magicmatatjahu:
@fmvilas Thanks for review! Could you check again?

src/components/Terminal/ProblemsTab.tsx
fmvilas: Captura de Pantalla 2022-11-15 a las 18 35 15

The default state doesn't make sense. It says "Show warnings" but I'm already seeing warnings 😅 I think, by default, all buttons should be checked.

magicmatatjahu: > I think, by default, all buttons should be checked.

It doesn't have a sense, because they people will need to uncheck three buttons if they will want to see only warnings. I changed state to show messages like:
- any button is checked: `Show only {diagnostic type}`
- one button is checked and we hover this checked button: `Show all diagnostics` and for other `Show {diagnostic type}`
- for rest: `Show {diagnostic type}` or `Hide {diagnostic type}`

WDYT? :) You can check in preview :)

fmvilas: Ok. Fair. It's weird that all buttons checked and all buttons unchecked will show you the same thing 😅

src/components/Terminal/ProblemsTab.tsx
```diff
@@ -1,73 +1,331 @@
...
+          </div>
+          <a href='https://stoplight.io/open-source/spectral' title='Spectral website' target='_blank' rel="noreferrer" className='text-white hover:text-blue-500 mx-2'>
+            <span>
+              Powered by
+            </span>
+            {' '}
+            <strong>
+              Spectral
+            </strong>
+          </a>
```
fmvilas, line 269: I'm curious. Why do we have this here? 🤔 It looks like an ad 😅

src/components/Terminal/ProblemsTab.tsx
magicmatatjahu: > Ok. Fair. It's weird that all buttons checked and all buttons unchecked will show you the same thing 😅

We should leave it as it's now, because people should have option to disable only one particular kind of diagnostics, when you have enabled others. The current solution isn't perfect but by changing it, people may have a worse UX.

> And it lets me disable warning, information, and hint messages. In such case, I think we should disable the buttons in this menu too and when you hover over them you'll see a tooltip saying it's been disabled in the settings and that you have to enable them back if you want to use them.

I will add.

src/components/Terminal/ProblemsTab.tsx
magicmatatjahu: Removed.

src/components/Terminal/ProblemsTab.tsx
magicmatatjahu: image

Can it be?

magicmatatjahu:
@fmvilas Could you make review again?

src/components/Terminal/ProblemsTab.tsx
fmvilas: Alright, this is looking much better now 👏

magicmatatjahu:
Finally 🚀 Thanks for everyone for review!

PR #456, opened 2022-10-13
fix: title in Kafka example

fmvilas:
Yeah, I don't think it needs explanation 😄


package.json
```diff
@@ -113,7 +113,7 @@
     "markdown-toc": "^1.2.0",
     "postcss": "^8.2.13",
     "raw-loader": "^4.0.2",
-    "react-scripts": "5.0.1",
+    "react-scripts": "4.0.3",
```
magicmatatjahu, line 116: Fixes https://github.com/asyncapi/studio/pull/451

fmvilas:
@magicmatatjahu why are you including all these commits in this PR? 🤔

magicmatatjahu:
@fmvilas

- we forgot to update version `2.4.0` -> `2.5.0`. It's fix so we can do it in one PR
- dependantbot updated `react-scripts` to the `^5` version and it breaks the build, so I revert to the `^4` version

magicmatatjahu (approved):

PR #468, opened 2022-10-21
fix: sonar codesmell Monaco Service

### LINKED ISSUE #447: Unnecessary assertion since it does not change the type of the expression.
More info on this issue can be found [here](https://sonarcloud.io/project/issues?resolved=false&id=asyncapi_studio&open=AYNkAOrH1pWMtoDLbyQC)

dhfherna:
Fix assertion unnecessary

- Modified monaco.service.ts

**Related issue(s)**
Resolve #447

### DIFF
--- a/src/services/monaco.service.ts
+++ b/src/services/monaco.service.ts
@@ -61,7 +61,7 @@ export class MonacoService {
   }
 
   static loadLanguageConfig(asyncAPIVersion: string) {
-    const monacoInstance = window.monaco;
+    const monacoInstance = window.monaco as any;
     if (!monacoInstance) return;
 
     const options = this.prepareLanguageConfig(asyncAPIVersion);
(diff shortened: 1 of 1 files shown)


src/services/monaco.service.ts
```diff
@@ -61,15 +61,15 @@ export class MonacoService {
   }
 
   static loadLanguageConfig(asyncAPIVersion: string) {
-    const monacoInstance = window.monaco;
+    const monacoInstance = window.monaco as any;
```
magicmatatjahu, line 64: TBH I don't like it, we should extend typings for `monacoInstance.languages` with adding `yaml` property with types like `monacoInstance.languages.json`. Could you do it? You should write types in https://github.com/asyncapi/studio/blob/master/src/react-app-env.d.ts

magicmatatjahu (changes requested):

PR #496, opened 2022-11-22
fix: change active/hover styles for navigation items

### LINKED ISSUE #193: Add tooltips to navigation and icons to new examples screen

### LINKED ISSUE #199: Accessibility: Make buttons on left nav bar accessible to screen readers
#### Reason/Context
Studio left nav bar

Current nav bar buttons don't provide any alternative text. This makes it hard to know what button does what.
The issue goes further because it means screen readers won't do their job, meaning Studio can't be used on all its potential by people with visual impairments. 

#### Description
I would suggest we, at least, add both `<title>` and `<desc>` elements to our inline svg's (the buttons are) so they act as alternative text.

More info here https://css-tricks.com/accessible-svgs/

magicmatatjahu:
**Description**

Styles are described in the https://github.com/asyncapi/studio/issues/199#issuecomment-1206690018

- change active/hover styles for navigation items
- add tooltip for settings modal

image

image

**Related issue(s)**
Resolves #193
Resolves #199


fmvilas (approved):
🚀

magicmatatjahu:
@fmvilas Thanks!

PR #497, opened 2022-11-22
fix: specify source when document is from the url

magicmatatjahu:
**Description**

Specify source when document is from the url - now all refs are resolved from url point of view.

cc @derberg 

**Related issue(s)**
Fixes https://github.com/asyncapi/studio/issues/392


derberg:
In UI it looks nice, clear info what file is rendered, and when I edit, localStorage is shown again. Thanks

side note, there is some other strange bug but you probably want to solve it separately. When I start editing, of course, local storage is used and all suddenly validation fails, because `base` changes. That is fine. But error about issue with dereferencing is not something I would expect: `TypeError: i.readFile is not a function`

magicmatatjahu:
@derberg I changed message to the:

image

Do you like it?

derberg:
Nice

small suggestion: if this is custom error for Studio only, then I suggest you write `File references are not yet supported in Studio` to make it clear that outside studio it works completely fine

src/services/specification.service.ts
```diff
@@ -34,12 +34,24 @@ export class SpecificationService {
...
 
     try {
-      const { document, diagnostics: validationDiagnostics, extras } = await parser.parse(rawSpec);
-      diagnostics = validationDiagnostics;
+      const { document, diagnostics: validationDiagnostics, extras } = await parser.parse(rawSpec, options);
+
+      // map messages of invalid ref to file
+      diagnostics = validationDiagnostics.map(d => {
+        if (d.code === 'invalid-ref' && d.message.endsWith('readFile is not a function')) {
+          d.message = 'File references are not yet supported';
```
magicmatatjahu, line 51:
```suggestion
          d.message = 'File references are not yet supported in Studio';
```

magicmatatjahu:
@derberg Done :)
@fmvilas Could you check?

fmvilas:
Is this problem known?

Captura de Pantalla 2022-11-24 a las 17 00 00

magicmatatjahu:
@fmvilas LOL, I haven't seen this error before. I tried replicating it locally and got nothing like that.... On the other hand, in the preview it occurs - presumably `monaco-yaml` diagnostic options can't download the schema from http... and `monaco-json` can.... I will create issue because it is not a serious problem, and autocompletion works, and it is the most important thing in json/yaml integration

Besides that, everything is ok in this PR?

magicmatatjahu:
@fmvilas https://github.com/asyncapi/studio/issues/509

fmvilas (approved):
Yeah, all good with this PR. This error is already happening in master so 👍

PR #500, opened 2022-11-24
refactor: simplify services

magicmatatjahu:
**Description**

At the moment every service is a static service. That PR creates singleton services which we can retrieve using global React context.

It's a part of refactoring which prepares the Studio for multiple files support. 

Next will be:
- simplify state management - move to the `zustand` library.
- simplify modal management.


fmvilas (approved):

magicmatatjahu:
@fmvilas I had to fix git conflicts. Could you check again?

fmvilas (approved):

PR #508, opened 2022-11-25
fix: does not fetch json-schema-07 from web.

magicmatatjahu:
**Description**

- does not fetch json-schema-07 from web - serialize the specification to have proper uri for `json-schema-draft-07` JSON Schema.

TESTING PR.

**Related issue**
Fixes https://github.com/asyncapi/studio/issues/509


magicmatatjahu:
@fmvilas I tried using the `http` protocol but it didn't work, then I tried removing the schema itself from our definitions and it still didn't work. Only after entering the schema as `https` (we have to do this manually and change every reference to `http`) does it work. Both validation and autocompletion. If you have another (worked) idea, you're welcome with contribution 😄

fmvilas:
So changing the `http` to `https` [here](https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.5.0.json) didn't work?

fmvilas:
Oof! 🤦 That's bad. I thought that JSON Schema was serving https://json-schema.org/draft-07/schema with the `"$id": "https://json-schema.org/draft-07/schema"` value. Would have been great if they could detect when HTTPS or HTTP is used and update accordingly. I know the URL shouldn't strictly be `$id` (and vice versa) but it would be very handy.

Pinging @relequestual.

fmvilas (approved):
Approving as this solution is better than what we have right now.

PR #516, opened 2022-11-29
fix: possibility of moving blocks in visualiser

magicmatatjahu:
**Description**

I found some weird bug: if we wanna move some block in visualiser, then in some cases it doesn't work and it can be possible to see "jumping" of the blocks between renders - from initial position to the destination. That PR fix that. Problem probably occurred through bumping the `react-flow-renderer` dep to the v11 version.

### DIFF
--- a/src/components/Visualiser/Controls.tsx
+++ b/src/components/Visualiser/Controls.tsx
@@ -25,7 +25,7 @@ export const Controls: FunctionComponent<ControlsProps> = () => {
   }, [animateNodes]);
 
   const reloadInterface = () => {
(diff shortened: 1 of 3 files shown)


fmvilas (approved):
Let's make sure we don't see the same problem in EDAVisualizer. Otherwise, when we switch to this package this will be lost.

magicmatatjahu:
@fmvilas There in EDAVisualiser are much bigger problems because it still uses v9 and not v11. I've already told Jonas that I'll have to change it and then we will merge integration. I will keep this strange bug in mind.

PR #520, opened 2022-12-01
refactor: remove old state management

magicmatatjahu:
**Description**

- remove whole old state management and replace in evert place with new one
- update unit tests

This is the last part of refactoring of the Studio to prepare app for multiple file supports.


fmvilas (approved):
Ok, blindly approving but let's get rid of this old state management. If some problem appears we can always fix it later 👍

PR #536, opened 2022-12-21
fix: comment mentioning the tutorial

starlightknown:
**Description**

- Refer - we could add a comment within the template that links to the tutorial. That way, users who accidentally run into this while using the Studio tool can immediately see we have a tutorial for learning how to debug their documents. https://github.com/asyncapi/website/pull/1022#issuecomment-1354032705

- also the link I have added is the link after the tutorial PR gets merged in future and not current deployed one as we have to change again
- ...

**Related issue(s)**


fmvilas (approved):
Good point.

magicmatatjahu:
@fmvilas we should wait for that PR 😅 https://github.com/asyncapi/website/pull/1022#issuecomment-1377150601 but ok.

PR #547, opened 2023-01-18
feat: highlighting nodes in flow diagram

### LINKED ISSUE #182: When clicking on items in the side bar (messages, operations, schema etc) be nice to highlight the nodes in the flow diagram
#### Reason/Context

At the moment when you have the sidebar and the documentation open, clicking on a message, operation etc will take you straight to the place in the docs.

I think it would be a nice touch to have the nodes in the flow diagram highlight (maybe blink animate) when clicking on these nav items.

So if the user has  this nav bar open....
image

Clicking on the items whilst looking at the flow diagram will highlight the nodes.

Ankitchaudharyy:
**Description**
Nodes will blink when `Servers`, `Operations`, and `Messages` are clicked. 

**Related issue(s)**
Fixes #182


magicmatatjahu:
@Ankitchaudharyy Thanks for contribution! When I click on server, the new styles of block is broken:

image

Also, "blink" effect is not good, we should change border of blocks and animate them or something like, not hide and show.

magicmatatjahu:
@Ankitchaudharyy Check this, is it better?

Ankitchaudharyy:
@magicmatatjahu, Looks good!

But I am still thinking, can we achieve the same output without using states?

magicmatatjahu:
@Ankitchaudharyy We need to animate some styles and using tailwind it's probably one possible solution. Handling normal DOm nodes in React context isn't good idea.

magicmatatjahu (approved):

PR #572, opened 2023-02-13
fix: update parser-js to latest version

magicmatatjahu:
**Description**

Update `parser-js` to latest version (`next-major.15`).

PR #615, opened 2023-05-02
fix: update parser-js to v2

smoya:
Replaces https://github.com/asyncapi/studio/pull/572

**Description**

This PR updates **@asyncapi/parser** to v2 and uses the recently released custom schema parsers.

Note: this PR does not change Studio to start using the new Parser-API but keeps using the old wrapper `const oldDocument = convertToOldAPI(document);`.

**Related issue(s)**
https://github.com/asyncapi/parser-js/issues/585

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -26,11 +26,11 @@
     "./LICENSE"
   ],
   "dependencies": {
-    "@asyncapi/avro-schema-parser": "^3.0.1",
+    "@asyncapi/avro-schema-parser": "^3.0.2",
     "@asyncapi/converter": "^1.2.0",
-    "@asyncapi/openapi-schema-parser": "^3.0.1",
-    "@asyncapi/parser": "^2.0.0-next-major.14",
-    "@asyncapi/react-component": "^1.0.0-next.47",
+    "@asyncapi/openapi-schema-parser": "^3.0.3",
+    "@asyncapi/parser": "^2.0.3",
+    "@asyncapi/react-component": "^1.0.0-next.48",
     "@asyncapi/specs": "^4.2.1",
     "@ebay/nice-modal-react": "^1.2.10",
     "@headlessui/react": "^1.7.4",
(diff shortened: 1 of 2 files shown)


src/services/parser.service.ts
```diff
@@ -28,8 +28,8 @@ export class ParserService extends AbstractService {
     });
 
     this.subscribeToFiles();
-    this.subscribeToSettings();
-    this.parseSavedDocuments();
+    await this.subscribeToSettings();
+    await this.parseSavedDocuments();
```
magicmatatjahu, line 32: It won't wait because subscription is a procedure, not function :) So I don't think so that this is needed.

src/index.tsx
```diff
@@ -32,7 +32,7 @@ function configureMonacoEnvironment() {
   };
 }
 
-async function bootstrap() {
+(async () => {
```
magicmatatjahu, line 35: Why did you change `bootstrap` function to something like that?

src/services/parser.service.ts
smoya: SonarCloud complains about this being missing.

src/index.tsx
smoya: SonarCloud complains (marks as a bug) with something like this iirc: "Promises must be awaited, end with a call to .catch, or end with a call to .then with a rejection handler"

magicmatatjahu: As we see, Sonarcloud was written by Java people 😆 Ok.

smoya:
> Something doesn't work. When I remove `info.title` from the spec then I see errors, but when I update the config to hide the warnings they aren't filtered (still with errors). Rest are good for me. Please only check that described bug, because it works as I described in current studio https://studio.asyncapi.com/

Being honest, I don't understand why changing the following lines back to what it was previously fixes the issue: https://github.com/asyncapi/studio/pull/615/commits/768b01c3aa0f8faab0498e813f566b824f53fe3e

PR #618, opened 2023-05-13
feat: add share functionality

### LINKED ISSUE #463: update url of studio after making changes
#### Reason/Context
It helps to share the document with others after someone makes changes to the document. Currently the link we share does not direct others to the same document and they still see the unchanged document.
- How will this change help?
This will make sharing easier with doing the base64 encoding and pasting it in url to make it work. 
- What is the motivation? 
This idea came when I was on call with lukasz and I am also working on a tutorial using studio.

Shurtu-gal:
**Description**

- A share functionality has been added to the toolbar.
- It converts the file contents to Base64 format for ease of sharing

**Related issue(s)**

Fixes #463 
Blocked by #807


fmvilas (commented):
I have concerns about building such long URLs. They may not work in all browsers. Not even in all the modern ones.

I just found the table below in the following Stack Overflow question: https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers.

```
Browser     Address bar   document.location
                          or anchor tag
------------------------------------------
Chrome          32779           >64k
Android          8192           >64k
Firefox          >64k           >64k
Safari           >64k           >64k
IE11             2047           5120
Edge 16          2047          10240
```

Can I ask you what's your use case? Why do you want to copy and share a URL but you don't want to copy and share the document itself? Honest question, just want to understand as I'm working on drafting the future of Studio and want to understand the needs better.

Shurtu-gal:
@fmvilas This was done as per this comment https://github.com/asyncapi/studio/issues/463#issuecomment-1289032605

Shurtu-gal:
If we don't want such long urls then we would have to use a storage. Either on our side or in their own Drives, dropboxes etc.

fmvilas:
> If we don't want such long urls then we would have to use a storage. Either on our side or in their own Drives, dropboxes etc.

Yes, that's in my plans for the next Studio version: https://github.com/orgs/asyncapi/projects/16/views/1.

---

Look, what I'd do is to add this feature knowing that it may fail in certain cases. I think we can live with that. Some people may experience issues but others (the majority?) will not so let's not block something that may be useful for them.

Are you ok with that @magicmatatjahu @derberg?

derberg:
I use this feature in mermaid all the time https://mermaid.live/
and yeah, some huge files might fail 🤷 but probably only for Edge and IE 😃

KhudaDad414:
@Shurtu-gal let's merge this PR. can you resolve the conflicts please?

Shurtu-gal:
@KhudaDad414 Done

Shurtu-gal:
@KhudaDad414 done

KhudaDad414 (approved):

KhudaDad414:
@jonaslagoni SonarCloud is complaining about the changes for #796 here. this is weird. 🤔

jonaslagoni:
Sonar cloud in a nut shell 🤷

Shurtu-gal:
> Sonar cloud in a nut shell 🤷

😆

jonaslagoni:
@KhudaDad414 I would just ignore sonar and force-push it manually.

Amzani:
Agree with @jonaslagoni we will anyway have duplicated code during the transition period to V3.

PR #639, opened 2023-06-03
fix: handle unexpected file formats

### LINKED ISSUE #623: Bug :  Restricting users from uploading any file format in the editor
#### Describe the bug
Currently, the editor allows users to upload any file format, which poses potential risks to the system's integrity and security. To ensure a safe and consistent user experience, it is necessary to restrict the file formats that can be uploaded in the editor.

#### How to Reproduce
1. Go to http://localhost:3000
2. Import file in editor.
3. Select any file .png, .gif etc.

- Screenshots
 
https://github.com/asyncapi/studio/assets/86047367/d1dc2fcc-176c-40dd-9929-c9914d9bf172

#### Expected behavior
The editor should enforce restrictions on the types of files that users can upload. On which file formats it support required maintainers suggestions. @magicmatatjahu @fmvilas @boyney123

Shurtu-gal:
**Description**

- I have added an accept attribute for better user experience by showing the files which are supported( currently JSON and YAML).
- Also there is a checker in import file which compares the MIME type as it cannot be handled in HTML.

**Related issue(s)**
Fixes #623

### DIFF
--- a/apps/studio/src/components/Editor/EditorDropdown.tsx
+++ b/apps/studio/src/components/Editor/EditorDropdown.tsx
@@ -39,6 +39,7 @@ export const EditorDropdown: React.FunctionComponent<EditorDropdownProps> = () =
     >
       <input
         type="file"
+        accept='.yaml, .yml, .json'
         style={{ position: 'fixed', top: '-100em' }}
         onChange={event => {
           toast.promise(editorSvc.importFile(event.target.files), {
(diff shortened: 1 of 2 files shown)


src/services/editor.service.tsx
```diff
@@ -131,6 +131,15 @@ export class EditorService extends AbstractService {
     if (!file) {
       return;
     }
+    
+    // Check if file is valid (only JSON and YAML are allowed currently) ----Change afterwards as per the requirement
+    if (
```
Amzani, line 136: Nice to have for future refactoring: We could have a constant file where we put theses kind of values, as it will be easier to manage by other contributors.

Shurtu-gal:
> @Shurtu-gal this PR has some conflicts

I have resolved them.

src/services/editor.service.tsx
Shurtu-gal: @Amzani Should I add a config file or sth else?

KhudaDad414 (approved):
LGTM

PR #680, opened 2023-06-20
ci: fix ci tests fail

KhudaDad414:
fixes tests fail due to craco cli being installed at project level https://github.com/asyncapi/studio/actions/runs/5313814389


package.json
```diff
@@ -13,6 +13,7 @@
   },
   "devDependencies": {
     "turbo": "^1.10.2",
+    "@craco/craco": "^7.1.0",
```
fmvilas, line 16: This seems incorrect to me. We only use craco for Studio. Shouldn't it be in the `apps/studio/package.json` file instead? 🤔

fmvilas (changes requested):

.github/workflows/if-nodejs-release.yml
```diff
@@ -38,24 +38,24 @@ jobs:
...
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: 18
           cache: 'npm'
           cache-dependency-path: '**/package-lock.json'
       - if: steps.packagejson.outputs.exists == 'true'
         name: Install dependencies
         run: npm install
```
fmvilas, line 55: I'm suspecting the problem is here. Maybe we should add `--workspaces` so it should become:

```suggestion
        run: npm install --workspaces
```

.github/workflows/if-nodejs-release.yml
```diff
@@ -38,24 +38,24 @@ jobs:
...
       - if: steps.packagejson.outputs.exists == 'true'
         name: Run test
-        run: npm test
+        run: npm test --if-present
```
fmvilas, line 58: Same as above. It should become `npm test --if-present --workspaces`.

KhudaDad414: I don't think we need the --workspaces flag here. as you can see in this run, everything works fine without it. https://github.com/KhudaDad414/studio/actions/runs/5333204486/workflow
it's the job of turborepo to handle workspaces here.

KhudaDad414: I even moved the `craco` package back to the `studio` app and everything seems to be working fine. The problem was with the version 14 of node that we were using. [turborepo min is 16](https://github.com/vercel/turbo/issues/2336) and we are using `tsup` which requires node 18.

fmvilas (approved):
Makes sense.

PR #681, opened 2023-06-20
feat(design-system): define basic design tokens

### LINKED ISSUE #666: Define the basic design tokens
We should define the basic design tokens for the design system. Things it should initially include:

- [ ] Typography
- [ ] Colors
- [ ] Spacing (margins and paddings)
- [ ] Borders
- [ ] Icons

Figma: https://www.figma.com/file/bPMB3lkMTOOMuKk0oGLuNm/Studio?type=design&node-id=0%3A1&t=5t1QnTliI6risXyA-1

Related to #636

fmvilas:
Fixes #666 

Preview URL: https://deploy-preview-681--asyncapi-studio-design-system.netlify.app/


apps/design-system/src/design-tokens/Borders.stories.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Border radius
+
+|Token|Description|
+|:--|:--|
+|`rounded-none`|(Default) No rounded border|
+|`rounded-sm`|2px radius|
+|`rounded`|4px radius|
+|`rounded-md`|6px radius|
+|`rounded-lg`|8px radius|
```
KhudaDad414, line 39: just curious, In some cases in the figma design, we have components with radios of 9px or spacing of 5px, should they follow the tokens or a new token be created for those spacing when need arise?

KhudaDad414 (approved):
🙇

apps/design-system/src/design-tokens/Borders.stories.mdx
fmvilas: I should either update the designs or we can create extra classes for these cases, we'll see. In any case, this is like a styleguide, they should not be added here just because one or two components have to use these styles. You see what I mean?

KhudaDad414: Got it. thanks.

PR #682, opened 2023-06-21
chore: remove package-lock.json files from apps + fix script

smoya:
**Description**

Turborepo seems to manage everything through one global `package-lock.json` file, the one located in the root of the repository. The rest are not needed.

This PR also fixes the `run studio` script which was running the design system instead.


KhudaDad414 (approved):
makes sense.

PR #688, opened 2023-06-22
fix: update Dockerfile to handle monorepo

KhudaDad414:
**Description**
- this PR updates the Dockerfile so we can generate a docker image for the release process.

resolves https://github.com/asyncapi/studio/actions/runs/5343506807/jobs/9686791298


fmvilas (approved):
I fully trust you here because I have no idea 😄

PR #691, opened 2023-06-23
feat: add Visual Editor/Code Editor Switch component

### LINKED ISSUE #673: Visual Editor / Code Editor Switch component
Implement the Switch component as seen in the [design](https://www.figma.com/file/bPMB3lkMTOOMuKk0oGLuNm/Studio?type=design&node-id=0%3A1&t=5t1QnTliI6risXyA-1).

Captura de pantalla 2023-06-16 a las 8 40 52

**Note:** Do not try to make it a generic Switch component. This one is a very special one with two icons so I would not call it `Switch` but, instead, `EditorSwitch` or something like that to avoid naming collisions with a future Switch component.

Related to #636

### LINKED ISSUE #692: Integrate eslint for style consistency
#### Reason/Context
Proposing the integration of ESLint into our project using the [existing configuration](https://github.com/asyncapi/studio/blob/master/apps/studio/.eslintrc) from the "studio" app.

KhudaDad414:
**Description**
* Add switch component.
* Add eslint script and config.
* Add `clean` script to remove all of the build and `node_modules` folders

**Related Issues**
Resolves #673
Resolves #692


packages/ui/components/EditorSwitch.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useState, useCallback } from 'react'
```
fmvilas, line 1: The component is not usable with the keyboard. My recommendation is that you wrap it with https://www.radix-ui.com/docs/primitives/components/switch.

fmvilas (changes requested):
Left a suggestion. Other than that, looks good 👍

.eslintignore
```diff
@@ -2,4 +2,5 @@ node_modules
 public
 docs
 lib
-build
\ No newline at end of file
+build
```
fmvilas, line 5: Why did you move this file and `.eslintrc.json` to the root? 🤔

package.json
```diff
@@ -5,9 +5,11 @@
     "url": "git://github.com/asyncapi/studio.git"
   },
   "scripts": {
+    "lint": "eslint",
```
fmvilas, line 8: Why do you need eslint here?

fmvilas (changes requested):

.eslintignore
KhudaDad414: because `.eslintrc.json` wasn't picked up by eslint and we couldn't enforce code and formatting consistency across the packages.

package.json
KhudaDad414: so that testing workflow can run the linter and ensure code consistency.

apps/design-system/src/components/OperationIcon.stories.tsx
```diff
@@ -1,14 +1,15 @@
 import { OperationIcon } from '@asyncapi/studio-ui';
 
-export default {
-    component: OperationIcon,
-    parameters: {
-        backgrounds: {
-          default: 'dark'
-        }
-    }
+const meta = {
```
fmvilas, line 3: What's the advantage of this change? 🤔 I honestly prefer the straight `export default` than having to define a variable and then export it.

.changeset/perfect-sloths-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@asyncapi/studio-ui": minor
+---
+
+Add EditorSwitch.
```
fmvilas, line 5:
```suggestion
Add EditorSwitch component.
```

fmvilas (changes requested):

apps/design-system/src/components/OperationIcon.stories.tsx
KhudaDad414: eslint recommends naming each export.
https://github.com/import-js/eslint-plugin-import/blob/v2.25.2/docs/rules/no-anonymous-default-export.md

I see we are not currently using that name in the  code but we might reuse some parts of the meta object in it's stories in other components. it would just add consistency between stories. but still, not a big deal 😆 I can disable this rule.

fmvilas: 🤷‍♂️😄

fmvilas (dismissed):

packages/ui/components/EditorSwitch.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import * as Switch from '@radix-ui/react-switch'
```
Amzani, line 2: @KhudaDad414 @fmvilas Radix-ui need an ADR ? or this was already decided in the past ?

packages/ui/components/EditorSwitch.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { EyeIcon, CodeBracketIcon } from '../components/icons'
+import classNames from 'classnames'
+
+// Define styles as constants outside the component to avoid recreating these during each render
+const transitionStyle = 'transition duration-200 ease-in-out'
+const switchClass = 'relative bg-gray-800 flex h-[35px] w-[62px] cursor-pointer rounded'
```
Amzani, line 8: Variable naming consistency, we are using both *Style and *Class.

packages/ui/components/EditorSwitch.tsx
KhudaDad414: I saw a few radix-ui components being used so I don't know 😆

fmvilas: I introduced it a few weeks ago so yeah, an ADR would be cool.

fmvilas (approved):

packages/ui/components/EditorSwitch.tsx
Amzani: Created an issue for this: https://github.com/asyncapi/studio/issues/734

Issue #663, opened 2023-06-15
Start using ADR (Architectural Decision Records)


fmvilas:
## Problem

> What's the story exposing the problem the users are facing

We'll have to make a lot of architectural decisions during the development of Studio. It would be cool to get a record of each of these decisions in a centralized place in this repo, so others can understand why we took some decisions in the past and also it would force us to not only think about the "what"s but also the "why"s of our decisions. This would prevent us from jumping straight to the latest cool technology without properly evaluating it before.

## Solution 

> Proposed solution with mockups, views...

I'm proposing we adopt ADR (Architectural Decision Records). Learn more about it here: https://betterprogramming.pub/the-ultimate-guide-to-architectural-decision-records-6d74fd3850ee.

Since the lifecycle of an ADR can change and there should be a way for people to discuss them before they're adopted, my proposal is that we keep them in Markdown format in a folder of this repo. Every time we want to add or change an ADR, we'll have to create a pull request and there we will be able to discuss the change via GitHub comments.

These pull requests should have a specific label (e.g., `kind/adr`) and their title should start with `[ADR-0001]` to easily identify which ADR is the PR about. A PR must not change multiple ADRs.

Lastly, given the label, title, and all the information about the PR, an automated script should take care of making sure we create valid ADR PRs.

Things to validate with the automation:

1. If a PR has the `kind/adr` label, its title has to start with `[ADR-xxxx]`.
2. If a PR title starts with `[ADR-xxxx]`, we should add the `kind/adr` label.
3. Once 1 and 2 are met, if a PR is changing multiple files, merge should be blocked.
4. Once 1 and 2 are met, if a PR is changing the file of another ADR, merge should be blocked.

## Rabbit holes

> Potential technical, design ... challenges

We already have automation enforcing a specific syntax for PR titles, i.e., `chore: `, `fix: `, `feat: `, etc. Not sure if we can skip this enforcement or if it would be easy.

My advice is, if that takes too long to change, then let's change the signature of the PR title to `chore: [ADR-xxxx]` or something like that. It's not beautiful but works.

## Scope

> Describe a list of tasks or issues that needs to be done ( you don't need to have an exhaustive list of all the tasks in the beginning) 

- [x] #676
- [x] #677
- [x] #678
- [x] #694
- [x] #705

## Out of bounds

> What won't be part of the scope

Do not try to come up with a UI to visualize the ADRs or anything like that. It's good enough to have them in a folder in Markdown format.

## Success criteria
[...]

PR #702, opened 2023-06-27
chore: [ADR-0001] use architecture decision records

### LINKED ISSUE #676: Create a Markdown template for the ADR
I'd suggest we start with something like this: https://github.com/joelparkerhenderson/architecture-decision-record/blob/main/templates/decision-record-template-by-michael-nygard/index.md.

Related to #663

Amzani:
**Description**

Start using Michael Nygard ADR template.

**Related issue(s)**

Resolves #676


README.md
```diff
@@ -50,3 +50,14 @@ npm run build:ds
 npm run build
 ```
 
+## Architecture decision records
+
+### Create a new architecture decision record
+
+- Install [adr tools](https://github.com/adr/adr-tools/blob/patch-1/INSTALL.md)
```
fmvilas, line 57: This repository has been archived in 2022. I think we should probably not use this tool.

fmvilas (changes requested):

README.md
Amzani: @fmvilas most of adr tools I find in the OSS are outdated I'm going to build a small one that implement this template.

fmvilas: I think a manual approach is fine. It's a really simple template :)

doc/adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
...
+
+Date: 2023-06-27
+
+## Status
+
+Accepted
+
+## Context
+
+We need to record the architectural decisions made on this project.
```
fmvilas, line 11: Can you expand on the why here? You can get it directly from https://github.com/asyncapi/studio/issues/663 if you don't want to scratch your head 😄

fmvilas (changes requested):

fmvilas (approved):

PR #706, opened 2023-06-27
chore: use changesets to manage release

### LINKED ISSUE #628: Need for urgent changes in GitHub Actions automation
This issue defines a list of tasks that need to be performed in this repo to make sure it's ci/cd automation works long term without any issues.

It is up to maintainers to decide if it must be addressed in one or multiple PRs.

Below are 3 different sections describing 3 different important ci/cd changes.

**IMPORTANT-START** 
For GitHub workflows that contain `This workflow is centrally managed in https://github.com/asyncapi/.github/` you do not have to perform any work. These workflows were already updated through the update in `.github`. The only exception is the workflows related to nodejs release. More details in **Upgrade Release pipeline - in case of nodejs projects** section
**IMPORTANT-END** 

### Deprecation of way data is shared between steps

Every single GitHub Action workflow that has `echo "::set-output name={name}::{value}"` need to be updated to follow `echo "{name}={value}" >> $GITHUB_OUTPUT`

We do not yet know when `set-output` will stop working. Previous disable date was 31.05 but now then say community needs more time.

[For more details read official article from GitHub](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

### Deprecation of node12

> 2nd bullet point is still relevant for you even if your projects in not nodejs project

- Every single workflow that uses [setup-node](https://github.com/actions/setup-node) action needs an update to follow `v3` version of this action, and make sure minimum node 14 is used
- Now this part is more complex. Problem with `node12` is that node-based GitHub Actions were using it in majority as a runtime environment. Look for example at [this action.yaml file for setup-node action v2](https://github.com/actions/setup-node/blob/v2/action.yml#L38). So the job that you have to do is go through all the workflows, and verify every single action that you use, make sure you are using the latest version that is not based on `node12`. I already did review a lot of actions as part of [this PR](https://github.com/asyncapi/.github/pull/226) so maybe you will find some actions there and can copy from me. For example `action/checkout` needs to be updated to v3.

Node12 end of support in action is probably September 27th.

[For more details read official article from GitHub](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)

### Upgrade Release pipeline - in case of nodejs projects

> ignore this section if your project is not nodejs project

You have 2 options. You can:

**A.** choose to switch to new release pipeline using instruction from https://github.com/asyncapi/.github/issues/205

**B.** stay with old release pipeline, and manually update GitHub workflows and actions used in it, you can inspire a lot from this PR https://github.com/asyncapi/.github/pull/226

I definitely recommend going with **A**

Workflows related to release:
- .github/workflows/if-nodejs-release.yml
- .github/workflows/if-nodejs-version-bump.yml
- .github/workflows/bump.yml

### LINKED ISSUE #651: Adapt Github Actions to make it work with monorepos
The release process that we have on AsyncAPI is meant to work with repositories containing a single application or package. Since [we want to make this repo a monorepo](https://github.com/asyncapi/studio/pull/638), we need to adapt the release process too.

### LINKED ISSUE #696: The release process releases version 1.0.0 for each app
#### Describe the bug
The release process released [version 1.0.0](https://github.com/asyncapi/studio/releases/tag/%40asyncapi%2Fstudio-v1.0.0) of the studio when there are no breaking changes and the previous version was `v0.18.1`.

the current version should have been `v0.19.0`.

KhudaDad414:
**Description**
Before merging we should make sure that all of the following things are done:

- [x] setup tags so the release process can pick em up.
- [x] disable version bump workflow.
- [x] check if there are any changes needs to be made as a result of not following semantic release.
- [x] disable  Conventional Commits workflow for PR titles.

**Related Issues**
Fixes #628 
Fixes #696
Fixes #651 
Related to #723 
**Testing**
Github Release: https://github.com/KhudaDad414/studio/


apps/studio/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/studio",
-  "version": "1.0.0",
+  "version": "0.0.0",
```
fmvilas, line 3: Why are we using 0.0.0 on Studio?

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "private": true,
+  "version": "0.18.1",
```
fmvilas, line 3: Shouldn't it be the version of Studio instead? I don't think we want to version the whole repo.

fmvilas (changes requested):

package.json
KhudaDad414: so we want to version only the studio app and ignore the changes that are being made against other packages and apps for now?

fmvilas: I'd say yes except for the UI package.

KhudaDad414: @fmvilas So I would know that I got it right this time:
1) we only make a release in GitHub and docker if changes are detected in the `ui` package or `studio` app.

3) the next release version would be `v0.19.0`

2) we track the version in the `studio`'s package.json

fmvilas: I think so yeah 😄

KhudaDad414:
@fmvilas @Amzani would love to know what you think.

.github/workflows/if-nodejs-release.yml
```diff
@@ -2,73 +2,18 @@
 # Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
 
 # It does magic only if there is package.json file in the root of the project
-name: Release - if Node project
+name: Release
```
fmvilas, line 5: If you're gonna change this, I'd also rename the file.

doc/adr/0002-use-changesets-for-publishing-and-versioning.md
```diff
@@ -0,0 +1,41 @@
+# 2: Use Changesets for Publishing Packages in the Monorepo
```
fmvilas, line 1: Don't add ADRs on the same PR. ADRs should have their own space where we track and discuss them, to keep things simple. Ideally, before we implement stuff.

fmvilas (changes requested):

doc/adr/0002-use-changesets-for-publishing-and-versioning.md
KhudaDad414: Raised a seperate PR here: https://github.com/asyncapi/studio/pull/723

fmvilas (dismissed):
LGTM! 🚀

KhudaDad414:
@fmvilas checks pass now. another review. 🙇

PR #711, opened 2023-07-02
feat: add slideover component

### LINKED ISSUE #674: SlideOver component
Implement the SlideOver component as seen in the [design](https://www.figma.com/file/bPMB3lkMTOOMuKk0oGLuNm/Studio?type=design&node-id=0%3A1&t=5t1QnTliI6risXyA-1).

Captura de pantalla 2023-06-16 a las 8 45 00

A slide-over component is a menu that flies on top of the existing UI, in this case, coming from the right side of the screen.

**Note:** This is not about implementing the content of the SlideOver component as you can see in the image above. It's about implementing a generic menu that can be closed (with ESC key too) and can contain anything we want. In other words, we pass `{children}` in React.

princerajpoot20:
**Description**

- Added SlideOver Component
Resolve #674


packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,46 @@
...
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key === 'Escape') {
+        onClose();
+      }
+    };
+    window.addEventListener('keydown', handleKeyDown);
+
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown);
```
fmvilas, line 26: We shouldn't register keydown events in the window object. We only want to close the SlideOver if, and only if, we press Escape while the SlideOver (or one of its children) has the focus.

packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,46 @@
...
+    };
+  }, [onClose]);
+
+  if (!isOpen) return null;
+
+  return ReactDOM.createPortal(
+    <div className="fixed inset-0 flex items-center justify-end z-50">
+      <div className="fixed inset-0 bg-black opacity-50" onClick={onClose} />
```
fmvilas, line 34: We shouldn't have a grey area like we usually have with modals. Check the design: https://www.figma.com/file/bPMB3lkMTOOMuKk0oGLuNm/Studio?type=design&node-id=96-2392&mode=design&t=vI18Txe7uMLMxePV-0

The SlideOver is intended to be present at the same as other content.

packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <div className="w-full max-w-lg bg-gray-950 text-gray-200 rounded-md shadow-lg overflow-auto relative p-2.5">
```
fmvilas, line 35: In the same way you have `onClick` in some elements, you can have `onKeyDown`: https://react.dev/reference/react-dom/components/common#keyboardevent-handler. Since it's a div, it may not allow you to register keydown by default but if you set the `tabIndex={0}` property on the div, it should become "focusable" and therefore able to capture key events.

packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <button className="absolute top-2 right-2" onClick={onClose}>
```
fmvilas, line 36: Add an aria-label for accessibility:

```suggestion
        <button className="absolute top-2 right-2" onClick={onClose} aria-label="Close">
```

Also, make sure it has a different style when the button has the focus.

packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,46 @@
...
+          <CloseIcon />
+        </button>
+        {children}
+      </div>
+    </div>,
+    document.body
```
fmvilas, line 42: Is there any reason you want to attach this to the document body directly instead of just rendering it as a normal component? 🤔

apps/design-system/src/components/SlideOver.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const handleOpen = () => setIsOpen(true);
+  const handleClose = () => setIsOpen(false);
+
+  return (
+    <div>
+      <button onClick={handleOpen}>Open SlideOver</button>
+
+      <SlideOver isOpen={isOpen} onClose={handleClose}>
+        <h2>Content of the SlideOver</h2>
+        {/* Other contents... */}
```
fmvilas, line 20:
```suggestion
```

No need to have comments here as they're not going to be shown.

fmvilas (changes requested):
That was fast! Left a few comments :)

princerajpoot20:
@fmvilas Done 🫡

packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    if (event.key === 'Escape') {
+      onClose();
+    }
+  };
+
+  return (
+    <div className="fixed inset-0 flex items-center justify-end z-50 backdrop-blur-[20px] opacity-[.67]">
+      <div className="w-full h-full max-w-lg bg-gray-950 text-gray-200 shadow-lg overflow-auto relative p-2.5 border-l-2 border-gray-700" tabIndex={0} onKeyDown={handleKeyDown}>
```
fmvilas, line 25: You don't need the parent div:

```suggestion
      <div className="w-full h-full max-w-lg bg-gray-950 text-gray-200 shadow-lg overflow-auto relative p-2.5 border-l-2 border-gray-700 backdrop-blur-[20px] opacity-[.67]" tabIndex={0} onKeyDown={handleKeyDown}>
```

Also, if we apply opacity here, the Close button and the content will be affected too. It shouldn't. Only the background should be affected by the opacity so I suggest that you apply it like this:

```suggestion
      <div className="w-full h-full max-w-lg bg-gray-950/[.67] text-gray-200 shadow-lg overflow-auto relative p-2.5 border-l-2 border-gray-700 backdrop-blur-[20px]" tabIndex={0} onKeyDown={handleKeyDown}>
```

See: https://tailwindcss.com/docs/background-color#changing-the-opacity

packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,35 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+  children: React.ReactNode;
+}
+
+const CloseIcon: React.FC = () => (
+  <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path d="M6 18L18 6M6 6L18 18" stroke="#94A3B8" strokeWidth="1.5" strokeLinecap="round" strokeLinejoin="round"/>
+  </svg>
+);
```
fmvilas, line 11:
```suggestion
```

You already have a close icon here: https://atoms.asyncapi.com/?path=/docs/design-tokens-icons--docs#icons. Just import it from `ui/icons`.

fmvilas (changes requested):

apps/design-system/src/components/SlideOver.stories.tsx
```diff
@@ -0,0 +1,23 @@
+import { useState } from "react";
+import { SlideOver } from "ui";
+
+export default {
+  component: SlideOver,
```
fmvilas, line 5:
```suggestion
  component: SlideOver,
  parameters: {
    layout: 'fullscreen',
    backgrounds: {
      default: 'dark'
    }
  },
```

packages/ui/components/SlideOver.tsx
```diff
@@ -0,0 +1,29 @@
...
+  if (!isOpen) return null;
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    if (event.key === 'Escape') {
+      onClose();
+    }
+  };
+
+  return (
+    <div className="fixed right-0 top-0 bottom-0 w-full max-w-lg h-full bg-gray-950/[.67] text-gray-200 shadow-lg overflow-auto p-2.5 border-l-2 border-gray-700 backdrop-blur-[20px]" tabIndex={0} onKeyDown={handleKeyDown}>
```
fmvilas, line 20:
```suggestion
    <div className="fixed right-0 top-0 bottom-0 w-full max-w-lg h-full bg-gray-950/[.67] text-gray-200 shadow-lg overflow-auto p-8 border-l-2 border-gray-700 backdrop-blur-[20px]" tabIndex={0} onKeyDown={handleKeyDown}>
```

fmvilas (changes requested):
Changed the padding to make it look like the design and made the story render as dark mode by default so we can properly see the border and everything.

fmvilas (approved):

fmvilas:
Congrats! Second PR in a week! 🙌

princerajpoot20:
@fmvilas It's all because of you. Thankyou for your support. 🙂

PR #714, opened 2023-07-04
fix: add missing arg to installer layer

### LINKED ISSUE #550: BASE_URL issue
When using the docker image with parameter `BASE_URL` it seem to not work as expected. Accessing the path it results in an `404` error.

```
docker run -it -p 8000:80 -e BASE_URL=/test asyncapi/studio:0.16.0
# open http://localhost:8000/test will result in 404
```

fgreinacher:
**Description**

The PR https://github.com/asyncapi/studio/pull/688/files changed the Dockerfile so that the `BASE_URL_PLACEHOLDER` arg was missing in the `installer` layer. Note that args need to specified for every layer.

**Related issue(s)**

Closes #550


fmvilas (approved):

fmvilas:
Thanks for the fix. We totally overlooked that 🙏

PR #723, opened 2023-07-05
chore: [ADR-002] use changesets to manage release

KhudaDad414:
Proposal to use changeset instead of semantic-release for versioning and publising packages to GitHub.


fmvilas (approved):
LGTM 👍

Issue #727, opened 2023-07-07
Setup NextJS app


Amzani:
Related to #725

Shurtu-gal:
/progress 90
Just some minor things left.

Shurtu-gal:
/progress 100
All things completed.

PR #737, opened 2023-07-13
add studio-next folder

fmvilas:
Related to #727 

The reason I'm pushing this empty folder is to set the stage so @Shurtu-gal and I can work in parallel.


fmvilas:
Nevermind, there's going to be no conflict since the folder name is the same 🤦

Issue #725, opened 2023-07-07
Studio New Architecture


Amzani:
## Problem

> What's the story exposing the problem the users are facing

As we are working on[ #684 ](https://github.com/asyncapi/studio/issues/684) we discovered some rabbit holes and challenges when it comes to NextJS migration.

To address theses challenges we want to setup up some architecture foundations for studio to enable a smooth and an easy migration to NextJS.

## Solution 

> Proposed solution with mockups, views...

This bet will cover 2 parts:

### First part

The idea is the have the new NextJS app inside the same mono-repo and reuse the existing components / ui design system. We want to avoid having long lived branch and avoid future conflicts.

Location of the new app 

```
apps/studio -> current app
apps/design-system
apps/studio-next -> NextJS app
```

The deployment and hosting should be accommodated to enable beta users to visite the studio-next app.  

### Second part

Create an ADR to decide the new NextJS app architecture and answer theses questions:

- Where to define the business logic ?
- What's the size of a component ?
- Architecture pattern to use (MVC?)
- State manager to use

## Rabbit holes

> Potential technical, design ... challenges

- Some components like monaco editor might be challenging to integrate.

## Scope

> Describe a list of tasks or issues that needs to be done ( you don't need to have an exhaustive list of all the tasks in the beginning) 

- [ ] #726
- [x] #727
- [x] #762

## Out of bounds

> What won't be part of the scope

We don't necessary need to have all the existing features up and running in the new app. This bet intention is to set the stage and enable future contributions.

## Success criteria
> How do we know we made a good bet

Amzani:
@fmvilas @Shurtu-gal don't hesitate to add anything missing here.

fmvilas:
I'd love to suggest that we get rid of Monaco and replace it with CodeMirror. Or at least try and see if it's better than what we have. Would you put it in a different bet issue?

fmvilas:
Yup, that definitely is the priority but don't want to lose the opportunity to kill Monaco with fire 🔥 😄

Shurtu-gal:
Funny thing, github itself uses code-mirror.
Meanwhile leetcode uses monaco along with next.

fmvilas:
Drafted some initial considerations for our archtiecture: https://github.com/asyncapi/studio/pull/739. Feel free to suggest adding, editing, or removing things.

Amzani:
Closing this for now! as studio will need a complete revamp/rewrite with the new DX Working group direction.

PR #739, opened 2023-07-13
add architecture considerations document

fmvilas:
**Description**

Some initial considerations for the new Studio architecture.

**Related issue(s)**
#725

### DIFF
--- a/apps/studio-next/docs/architecture.md
+++ b/apps/studio-next/docs/architecture.md
@@ -1,0 +1,156 @@
+# Studio Architecture Considerations
+
+Studio is an isomorphic application built on top of the [Next.js](nextjs) framework.
+
+## Routing
+
+Routing is handled by the Next.js framework. Please have a look at [the documentation](https://nextjs.org/docs/app/building-your-application/routing) to better understand how it works.
+
+As a rule of thumb, every UI state that could be worth sharing with others, deserves its own URL and therefore its own route.
+
+## Separation of concerns
+
+### Components
+
+> **Warning:** We should never ever put business logic on components.
+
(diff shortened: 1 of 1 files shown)


Shurtu-gal:
@fmvilas The problems faced using the current services may also be mentioned.
Going forward, what I propose to do is :- 
 - Hoist shared components such as editor to context and pass them onto hooks [example](https://github.com/Shurtu-gal/studio/blob/feat/migrate-next/apps/studio/src/hooks/useServices.tsx)
 - Use hooks to share the functions and store application logic. [example](https://github.com/Shurtu-gal/studio/tree/feat/migrate-next/apps/studio/src/hooks)

This works well currently; even things such as navigation in the editor, template and visualiser are working correctly.

fmvilas:
@Shurtu-gal I wonder if that's something related to the current architecture. Maybe, with the new architecture, we can find a way to make this easier? I mean, with a couple of exceptions, everything should be executable on the server and on the client. What do you think?

Shurtu-gal:
Yeah, if we separate it instead of maintaining it on the client. These problems will be solved by themselves. This way, the object of concern may be separated, and services may be exposed as APIs. 

P.S.:- But things like navigation and theme setting would have to be handled on the client side using context and hooks.

Amzani:
@fmvilas @Shurtu-gal I think it's a valid question that we should cover under the new architecture ? **When do use Hooks ?** and also **How to share state between components ? do we use React Context or Zustand**

fmvilas:
Agree. We'll have to address all these questions. Who wants to take the lead here?

Amzani:
Still relevant

PR #768, opened 2023-08-10
feat: update Parser-JS to v2.1.0-next-major-spec.5 so it really enables v3 validation

smoya:
**Description**

This PR updates the `@asyncapi/parser` package to the latest version of `next-major-spec` branch so it enables (really) the validation for AsyncAPI v3 documents.

cc @Amzani


smoya:
Don't get why [this](https://github.com/asyncapi/studio/pull/768#issuecomment-1672867152) is happening about changeset 🤷 ? It's just a version bump of the parser...

magicmatatjahu (approved):

PR #784, opened 2023-09-06
add AppCard component

KhudaDad414:
**Description**

- ...
- ...
- ...

**Related issue(s)**


packages/ui/components/AppCard.tsx
```diff
@@ -0,0 +1,33 @@
...
+    <>
+      <div
+        className={`bg-gray-800 border-gray-800 rounded-lg w-[523px] border-2 ${className} ${
+          isActive ? ' border-pink-800/30 shadow-service-card' : ''
+        }`}
+      >
+        <div className="flex flex-col gap-2 px-5 py-3">
+          <h3 className="text-base font-medium text-gray-100">{name}</h3>
+          <div className='flex gap-1'>
+            {protocols.map((protocol, index) => (<ProtocolBadge protocol={protocol} key={index} />))}
```
fmvilas, line 22: We should also show if the app is a client or a server (or both). These icons:

Captura de pantalla 2023-09-06 a las 17 16 23

From left to right, these are Server and Client respectively.

So you should probably take two more props in this component: `isClient` and `isServer`. I'd put them in the beginning before the protocols.

packages/ui/tailwind.config.js
```diff
@@ -1,6 +1,13 @@
-const sharedConfig = require('tailwind-config/tailwind.config.js');
+import sharedConfig from 'tailwind-config/tailwind.config.js';
 
 module.exports = {
   content: ['./components/**/*.tsx'],
   presets: [sharedConfig],
+  theme: {
+    extend: {
+      boxShadow: {
+        'service-card': '0px 0px 29px 0px rgba(190, 24, 93, 0.50)',
```
fmvilas, line 9: Do not tie CSS styles to its usage. Instead, use a name that represents what the styles do/are:

```suggestion
        'active-shadow': '0px 0px 29px 0px rgba(190, 24, 93, 0.50)',
```

fmvilas (changes requested):
Lovely! Looks like in the design 🚀 

Just left some comments.

KhudaDad414:
@fmvilas another round of review 🙇

fmvilas:
Badges separation between client and server, and the protocols are not the same. Please make it the same 🙏 

Captura de pantalla 2023-10-10 a las 13 00 51

KhudaDad414:
@fmvilas done. but left the space between client/server badge and protocol badges to keep them separated.

Amzani (approved):
LGTM

packages/ui/components/AppCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+  badges: Info[];
+  isClient: boolean;
+  isServer: boolean;
+  className?: string;
+}
+
+export const AppCard = ({isActive = false, name, description, badges, className, isClient, isServer}:AppCardProps) => {
+  return (
+    <>
+      <div
```
princerajpoot20, line 16:
```suggestion
      <div
        aria-label={`${name} Application Card, ${isActive ? 'active' : 'inactive'}`}
```
@KhudaDad414 Could we also add aria-label for accessibility.

packages/ui/components/AppCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+        className={`bg-gray-800 border-gray-800 rounded-lg w-[523px] border-2 ${className} ${
```
princerajpoot20, line 17:
```suggestion
        className={`bg-gray-800 border-gray-800 rounded-lg max-w-[523px] w-full border-2 ${className} ${
```
We can make it responsive for smaller screens.

packages/ui/components/AppCard.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <div className="flex flex-col gap-2 px-5 py-3">
+          <h3 className="text-base font-medium text-gray-100">{name}</h3>
+          <div className='flex gap-1'>
+            { (isClient || isServer) && 
+            <div className='mr-2 flex gap-1'>
+              {isClient && <ServiceInfoBadge info='client'/>}
+              {isServer && <ServiceInfoBadge info='server'/>}
+            </div>
+            }
+            {badges.map((badge, index) => (<ServiceInfoBadge info={badge} key={index} />))}
```
princerajpoot20, line 30:
```suggestion
            {Array.from(new Set(badges)).map((badge, index) => (<ServiceInfoBadge info={badge} key={badge + index} aria-label={`Badge for ${badge}`} />))}
```

- How about we add a condition to remove duplicate badges? This way, we'll only display unique badges.

princerajpoot20, line 30:
```suggestion
            {Array.from(new Set(badges))
              .filter(badge => validBadges.includes(badge))
              .map((badge, index) => (
                <ServiceInfoBadge info={badge} key={badge + index} aria-label={`Badge for ${badge}`} />))}
```
- I noticed that invalid badges leave a blank space (reserved for that invalid badge). 
What do you think about adding a check for this also? If a badge is not valid, we could simply ignore it instead of reserving blank space.

For this, we need to add an export of the list of valid badges from `ServiceInfoBadge`
```
export const validBadges: Info[] = ['http', 'kafka', 'websocket', 'amqp', 'mqtt', 'googlepubsub', 'ibmmq', 'nats', 'pulsar', 'redis', 'sns', 'sqs', 'solace', 'stomp', 'client', 'server'];
```

princerajpoot20 (changes requested):
Great work @KhudaDad414! Just a few suggestions to consider.

packages/ui/components/AppCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <div className="flex flex-col gap-2 px-5 py-3">
+          <h3 className="text-base font-medium text-gray-100">{name}</h3>
+          <div className='flex gap-1'>
+            { (isClient || isServer) && 
+            <div className='mr-2 flex gap-1'>
+              {isClient && <ServiceInfoBadge info='client'/>}
+              {isServer && <ServiceInfoBadge info='server'/>}
+            </div>
+            }
+            {Array.from(new Set(badges)).map((badge, index) => (<ServiceInfoBadge info={badge} key={badge + index} aria-label={`Badge for ${badge}`} />))}
```
fmvilas, line 31: I would extract this to a variable so it's easier to understand. Above the `return` on line 14, I would add this:

```js
const dedupedListOfBadges = Array.from(new Set(badges)).map((badge, index) => (<ServiceInfoBadge info={badge} key={badge + index} aria-label={`Badge for ${badge}`} />))
```

And then in this line I would just add `{dedupedListOfBadges}`. The code would be much more readable IMHO.

fmvilas: Also, for the `aria-label`, this is what a screen reader would say so imagine having 4 or 5 badges. The screen reader would say: "Badge for client, Badge for server, Badge for websocket, Badge for HTTP, Badge for kafka" 😅 Badge badge badge 😂 

I suggest that we improve the root element `aria-label` with a better explanation and would mark all these badges with `aria-hidden={true}`. This way we can create a meaningful sentence like "This application is both a server and a client. It uses the WebSocket, HTTP, and Kafka protocols".

```suggestion
            {Array.from(new Set(badges)).map((badge, index) => (<ServiceInfoBadge info={badge} key={badge + index} aria-hidden={true} />))}
```

Then on line 17, we change it to:

```html
aria-label={ariaLabel}
```

Where `ariaLabel` is something like "{name}. This application is active/inactive, and it's both a server and a client. It uses the WebSocket, HTTP, and Kafka protocols."

Note we're missing the description though. I'm not sure if a screen reader would read the paragraph below because it's not focusable. We should add https://storybook.js.org/addons/addon-screen-reader to Storybook so we can properly test what our users will experience using a screen reader.

fmvilas (changes requested):

packages/ui/components/AppCard.tsx
princerajpoot20: > I would extract this to a variable so it's easier to understand. 

Agreed, extracting the badges list makes the code cleaner.

> Also, for the aria-label, this is what a screen reader would say so imagine having 4 or 5 badges. The screen reader would say: "Badge for client, Badge for server, Badge for websocket, Badge for HTTP, Badge for kafka" 

I see your point, and I agree. Right now, it is not a fit for 4-5 badges.

> I suggest that we improve the root element aria-label with a better explanation and would mark all these badges with aria-hidden={true}. This way we can create a meaningful sentence like "This application is both a server and a client. It uses the WebSocket, HTTP, and Kafka protocols".
We should add https://storybook.js.org/addons/addon-screen-reader to Storybook so we can properly test what our users will experience using a screen reader.

Sounds perfect.

packages/ui/components/AppCard.tsx
KhudaDad414: we are already marking it's type as `Info[]` so it will give error in Typescript. so I am more in favour of letting typescript give the error.

princerajpoot20: Sorry, I didn't get this, could you please explain that further.

KhudaDad414: @princerajpoot20 I mean Typescript doesn't allow invalid badges:

Screenshot 2023-11-29 at 16 33 31

princerajpoot20: @KhudaDad414 Okay, got it. Makes sense. 👍

packages/ui/components/AppCard.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const AppCard = ({isActive = false, name, description, badges, className, isClient, isServer}:AppCardProps) => {
+  const dedupedListOfBadges = Array.from(new Set(badges)).map((badge, index) => (<ServiceInfoBadge info={badge} key={badge + index} aria-hidden={true} />))
+  
+  const ariaDescriptionParts = [];
+  if (isClient) ariaDescriptionParts.push('client');
+  if (isServer) ariaDescriptionParts.push('server');
+  const protocols = badges.map(badge => badge.toLowerCase()).join(', ');
+  if (protocols) ariaDescriptionParts.push(`uses the ${protocols} protocols`);
```
fmvilas, line 20:
```suggestion
  if (protocols) ariaDescriptionParts.push(`uses the following protocols: ${protocols}`);
```

I think this way it will sound more natural.

fmvilas (changes requested):

fmvilas (approved):
LGTM 👍

PR #792, opened 2023-09-16
feat: add client and server icon

### LINKED ISSUE #791: Missing Client and Server Icon.
### Problem

Missing Client and Server Icon.

### Solution

Add the missing Client and Server Icon and rename ProtocolBadge to ServiceInfoBadge

- [x] Add Client Icon
- [x] Add Server Icon
- [x] Rename ProtocolBadge to ServiceInfoBadge

princerajpoot20:
Resolves #791 Added the missing Client and Server Icon and rename ProtocolBadge to ServiceInfoBadge


apps/design-system/src/components/ServiceInfoBadge.stories.tsx
```diff
@@ -16,6 +16,18 @@ export const HTTP = {
   }
 }
 
+export const Server = {
+  args: {
+    protocol: 'server'
```
fmvilas, line 21: We should probably rename `protocol` to `info` or something similar since it's not a Protocol Badge anymore. What do you think?

fmvilas (changes requested):
Just a small request 👍

apps/design-system/src/components/ServiceInfoBadge.stories.tsx
princerajpoot20: `info` would be a good choice, or we could use `serviceName` or `serviceInfo`, which would be a bit clearer.

princerajpoot20: Sure :)

fmvilas (approved):
Awesome! 🚀

Issue #803, opened 2023-10-04
Studio breaks when trying to generate code/document


AceTheCreator:
when trying to generate the docs/code, some errors appear in the console
![image (1)](https://github.com/asyncapi/studio/assets/40604284/fb648e8b-3e0a-4762-8301-53f1c7591e6b)

Amzani:
@AceTheCreator could you provide some more details ? what are the steps you did to have theses errors and also you browser details.

Amzani:
Duplicate of [#926](https://github.com/asyncapi/studio/issues/926)

PR #804, opened 2023-10-04
fix: studio breaking when trying to generate code/document

AceTheCreator:
This PR fixes the following issue. #803 

cc @fmvilas @Souvikns


apps/studio/src/components/Modals/Generator/TemplateParameters.tsx
```diff
@@ -106,7 +106,7 @@ export const TemplateParametersSans: ForwardRefRenderFunction<TemplateParameters
     const servers = document?.servers();
     const availableServers: string[] = [];
     Object.entries(servers || {}).forEach(([serverName, server]) => {
-      if (supportedProtocols.includes(server.protocol())) availableServers.push(serverName);
+      if (server.protocol && supportedProtocols.includes(server.protocol())) availableServers.push(serverName);
```
fmvilas, line 109: How can it be that `protocol` is not a function? What is it instead? 🤔

fmvilas:
Also, if you plan to release a new version, please add a changeset: https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

apps/studio/src/components/Modals/Generator/TemplateParameters.tsx
AceTheCreator: So, based on what I noticed during the iteration, some index of server in `server.protocol` is returning `undefined`

AceTheCreator:
> Also, if you plan to release a new version, please add a changeset: https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

Why should it be a new version if I'm just fixing a tiny bug 🤔

KhudaDad414:
@AceTheCreator it is a bug fix so it should release a patch version.

KhudaDad414 (approved):
LGTM!

AceTheCreator:
@fmvilas, can you approve this?

PR #807, opened 2023-10-05
resolve code duplicate issue

KhudaDad414:
**Description**
SonarClould check fails for the duplicate code in the open PRs:
- https://github.com/asyncapi/studio/pull/784/checks?check_run_id=17432376236
- https://github.com/asyncapi/studio/pull/618#issuecomment-1748713981


KhudaDad414:
@fmvilas 👁️ 🙏

apps/studio/src/components/Navigation.tsx
```diff
@@ -67,6 +68,142 @@ const ServersNavigation: React.FunctionComponent<NavigationSectionProps> = ({
...
+        }
+      >
+        Channels
+      </div>
+      <ul>{channels}</ul>
+    </>
+  );
+};
+
+const V3OperationsNavigation: React.FunctionComponent<NavigationSectionProps> = ({
```
Amzani, line 123: @KhudaDad414 we will have more and more duplicated code as we are in the transition period to V3. 
We should just add an exception in SonarCloud, and clean after ending support for V2.
Not sure if there is an other clean way to do it.

Amzani, line 123: Naming suggestion, I find `OperationNavigationV3` more readable than `V3OperationsNavigation`

Amzani (changes requested):

KhudaDad414:
#821 resolves the issue.

PR #808, opened 2023-10-05
Add ChipInput component

### LINKED ISSUE #753: Visual Designer: ChipInput component
> ChipInput allows the users to enter text that can be converted to chips.

Screenshot 2023-07-31 at 15 00 34

Related to #748

princerajpoot20:
**Related issue(s)**

Resolves #753


packages/ui/components/ChipInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+      setInputValue('');
+    }
+  };
+
+  const handleDelete = (chipToDelete: string) => () => {
+    setChips((chips) => chips.filter((chip) => chip !== chipToDelete));
+  };
+
+  return (
+    <div className="flex flex-wrap items-center p-2 bg-gray-900 rounded">
```
KhudaDad414, line 27: based on the design, this component should have a border as well.

KhudaDad414 (changes requested):
left some comments.

princerajpoot20:
@KhudaDad414 Thankyou for the review. I have made the changes 🙂

packages/ui/components/ChipInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+      ))}
+      <input
+        name={name}
+        id={id}
+        type="text"
+        onKeyDown={handleKeyDown}
+        className="p-1 bg-gray-900 text-white rounded outline-none"
+        placeholder={placeholder}
+        disabled={isDisabled}
+        defaultValue={'registr'}
```
fmvilas, line 50: I would not default it to this but an empty string. You can then create multiple cases in the design system:

1. How it looks with all the defaults
2. How it looks with some default text
3. How it looks with a custom placeholder
4. And so on...

fmvilas (changes requested):
Left some comments.

I also noticed that if I delete the text and press delete again, it doesn't remove the last chip. I was expecting to be able to remove chips with the keyboard.

Also, I should be able to navigate through the chips using the keyboard (tab key) and remove a chip if needed (probably the delete key).

Hope that makes sense. It's looking awesome so far ♥️

princerajpoot20:
@fmvilas Thank you for the review. 🙂 

> Hope that makes sense. It's looking awesome so far ♥️

Definitely makes sense. My bad. I missed these things. 
I have passed the remaining values from the design system but forgot to pass `defaultValue` from there. Instead, I just hardcoded it in the `ChipInput.tsx` 😄.

About navigating through chips, it's a great point that I missed out on. I use tab a lot to navigate. it just makes it so convenient. 
I have used `firstChipRef` to get a reference to the first chip. So, when we are in the `InputTestField` and press tab, navigation will start from the `first chip` and continue from there.

fmvilas (approved):
This is beautiful 👏 

Great stuff! 🙌

KhudaDad414 (approved):
LGMT!

PR #848, opened 2023-11-14
fix: update official invalid example to match tutorial

derberg:
making sure invalid example match the one mentioned in tutorial 

https://v3.asyncapi.com/docs/tutorials/studio-document-validation

marked as `fix` as I kinda assume it will be automatically released once I do it


KhudaDad414 (approved):
LGTM!

Issue #979, opened 2024-02-27
Disable the generators that are not yet supported in V3


Amzani:
Some generators don't support yet V3

- @asyncapi/dotnet-nats-template
- @asyncapi/ts-nats-template
- @asyncapi/python-paho-template
- @asyncapi/nodejs-ws-template
- @asyncapi/java-spring-cloud-stream-template
- @asyncapi/go-watermill-template
- @asyncapi/java-spring-template
- @asyncapi/nodejs-template
- @asyncapi/java-template
- @asyncapi/php-template

## Current behavior

[...]

PR #982, opened 2024-02-27
fix: Disable unsupported generators for AsyncAPI V3 (#979)

kalpesh-d:
**Description**

- Updated the GeneratorModal component to disable unsupported generators for AsyncAPI V3.
- Added a warning message for unsupported generators with a link to the tracking page.

**Related issue(s)**
Fixes #979

PR #984, opened 2024-02-29
fix: Disable the generators that are not yet supported in V3

riyanrathore26:
Disable the generators that are not yet supported in V3

fix: #979 

asy


Amzani:
Sorry @riyanrathore26 there is already a pending [PR](https://github.com/asyncapi/studio/pull/982/files) fixing #979 
Looking forward to the next one.
Please make sure you comment on issues before submitting PRs
Thanks

Issue #684, opened 2023-06-22
Minimal studio App with a working editor and AsyncAPI documentation


fmvilas:
As part of the discussion in #661, I'm creating this issue to track the migration since it's a huge task.

Related to #661

Amzani:
/progress 10

This is a test

Shurtu-gal:
/progress 20
Figuring out how to add Monaco editor in next.js.

Shurtu-gal:
/progress 40
I've finished the initial setup with the Monaco editor and got a basic skeleton up and running.
Next, got to figure out how to incorporate services.

Amzani:
still relevant

Shurtu-gal:
/progress 40

Some things to note:-
- We are migrating to codemirror now.
- Although lightweight it has many features ( in the form of extensions and plugins which can be added on top of it )
- Would be adding the ADR soon

Things left:-
- Adding menu and AsyncAPI docs

KhudaDad414:
/progress 90

PR is ready for review: https://github.com/asyncapi/studio/pull/1062

Some known issues with current `studio-next` app:
* There is no docker container for it yet.
* tests haven't been ported.
* code generation doesn't work.
* Uses Monaco instead of CodeMirror.

PR #701, opened 2023-06-27
feat: migrate to next

Shurtu-gal:
#### Description:- 
This is made to migrate the current react framework to NextJs 13

- [x] Install Basic packages for next and initialise it 
- [x] Add tailwind support
- [x] Setup web pack
- [ ] Setup services correctly to run on the server,
- [x] Add support for MonacoEditor
- [ ] Add loader for YAML and JSON files.
- [ ] Clean up the packages
- [ ] Update the actions
- [ ] Write some basic tests.

#### Related issues:-
Resolves #684

Issue #224, opened 2022-01-09
Nice to have: Open Graph link preview image according to the document to open


smoya:
#### Reason/Context
Thanks to the `?url=<url-of-file>` and `?base64=<base64-encoded-doc>` query param, Studio can load most of files (yes, not all of them, see https://github.com/asyncapi/studio/issues/127). I expect users will use that to share their AsyncAPI docs.

Whenever a link to Studio (with or without those query params) is pasted into social media (Twitter, Linkedin, Facebook, Slack...), the preview image is this one: 

![image](https://user-images.githubusercontent.com/1083296/148680670-98b88679-c2b4-449b-a671-d03aa06c1f83.png)

It is a great pic, however it says nothing about the file being shared.

What if we could dynamically generate the preview image based on the file being shared? For example, the title, description and some stats could be shown.

#### Description

I created a POC based on https://github.com/vercel/og-image, available in [my fork](https://github.com/smoya/og-image) which is a server that generates dynamic images for being used on [Open Graph](https://ogp.me/) image meta tags. This works by generating a dynamic HTML, making an screenshot of it through headless Chromium, and serving the resulting image.

The server accepts a `?base64=<base64-encoded-doc>` query param, and generates an image that contains the AsyncAPI doc Title, Description, number of servers, channels and messages.

Despite the horrible design, the service is able to generate the following:

Based on the following AsyncAPI doc:

```yaml
asyncapi: '2.2.0'
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    subscribe:
      message:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
```
[...]

PR #1026, opened 2024-03-18
feat: initial try to get the parser working

Shurtu-gal:
**Description**

- Initial Implementation for docs

**Related issue(s)**
#684


KhudaDad414:
@Shurtu-gal Hey man, I want to be a part of this initiative and provide some help if I can. For that, I tried to migrate the studio to Next.js before and faced some challenges but overall it was smooth.

But I had to have `"use client"`on my main page and it was working as a SPA without any features of Next.js in it.

Can you provide what the challenge is here? And how you are trying to resolve it? 🙏

Shurtu-gal:
@KhudaDad414 this is the error which I have been facing for the last couple of days. 
https://app.netlify.com/sites/studio-next/deploys/65f8a395c20fb40008bfa630#L210-L233

KhudaDad414:
@Shurtu-gal the error seems... strange to say the least.
I have uploaded my version of the studio in next here: https://github.com/KhudaDad414/studio/tree/playground-next

it works great in dev and builds fine. There may be some rough edges that we need to iron out.

Shurtu-gal:
> But I had to have "use client"on my main page and it was working as a SPA without any features of Next.js in it

@KhudaDad414 then is there any benefit in shifting to Next.js then 🤔

Shurtu-gal:
I believe there was a misunderstanding. It will be like `AsyncAPI doc` in the browser but will get parsed over at the server. 

However, while writing this I can think of the disadvantages as well, as there will be too many calls to the API functions.

Amzani:
@Shurtu-gal, @KhudaDad414 there are no immediate benefits, but this will unlock a couple of features:

- Completely remove `server-api` dependency and make the code/doc generation/conversion on the server side.
- Enable server side plugins  (auth...)
- Enable APIs: (e.g. Synchronisation with IDEs extensions...)
- https://github.com/asyncapi/studio/issues/224
- Potential performance improvement

If we have an alternative to `nextjs` to enable that I'm fine with it

Shurtu-gal:
So should I move forward with using 'use client' everywhere?

But I believe this will cause problems down the line 😔.

Maybe 🤔 we could resolve this, let me debug it a little

Also @KhudaDad414 @Amzani your help and inputs would be great too.

KhudaDad414:
@Shurtu-gal I am afraid there is no other way.

Amzani:
@Shurtu-gal @KhudaDad414 what if we use `use client` at layout level ? so we don't need to repeat it in each component.

Shurtu-gal:
> @Shurtu-gal @KhudaDad414 what if we use `use client` at layout level ? so we don't need to repeat it in each component.

That would work, let me try it out and I will let you know.

Amzani:
@Shurtu-gal @KhudaDad414 NextJS might sound useless in the beginning, especially as the components are tightly coupled to Browser API, for instance, the TopBar might directly use browser Local storage, so we can't make it a Server Side component.

Once we complete the migration we can follow this pattern that @fmvilas defined in the following: https://github.com/asyncapi/studio/pull/739/files (this is still relevant), so a component like TopBar could be a server-side component for instance.

Shurtu-gal:
> > @Shurtu-gal @KhudaDad414 what if we use `use client` at layout level ? so we don't need to repeat it in each component.
> 
> That would work, let me try it out and I will let you know.

I tried that, but it is showing the same errors. Furthermore @KhudaDad414 your implementation is also not working. FYI there was a similar try in this direction before too in this [PR](https://github.com/asyncapi/studio/pull/701).

KhudaDad414:
@Shurtu-gal let me give it a try. 🤔

KhudaDad414:
Hey @Shurtu-gal, I have opened [a PR](https://github.com/asyncapi/studio/pull/1062) with a solution. (there were too much complications pushing the changes to this PR)
please let me know what you think.

Shurtu-gal:
@KhudaDad414 I am fine with it. @Amzani I guess as there are not many changes to the original app, we should wait before we migrate to CodeMirror. Or I can do it again after the migration PR gets merged.

Closing this PR then.

PR #1060, opened 2024-04-03
[ADR-0006] Use PNPM package manager

KhudaDad414:
**Description**
- I propose we use pnpm instead of npm in studio


Shurtu-gal:
> We can add the `pnpm-lock.yaml` and `pnpm-workspace_yaml` as well.

The README should also be changed.

KhudaDad414:
@Shurtu-gal after we have an approval from other code-owners, I can move forward with an another PR implementing the change and making sure it works properly with Turborepo.

Shurtu-gal (approved):
I myself have been using `pnpm` for quite some time now, and agree with this change 🥳.

Amzani (approved):
LGTM 🚀

Issue #661, opened 2023-06-15
Start using a React framework


fmvilas:
## Problem

> What's the story exposing the problem the users are facing

CRA (create-react-app) is dead. [They're gonna replace it with a launcher where you can select a framework](https://github.com/reactjs/react.dev/pull/5487#issuecomment-1409720741).

## Solution 

> Proposed solution with mockups, views...

That said, we should evaluate a framework and bet on it (no pun intended 😄). Things we need to take into account at first sight:

1. The framework needs to be well maintained and have an active community behind it. We don't to repeat this in a few months 😅 
2. It needs to be possible to be deployed to Netlify. We have an open-source plan with them and that means we won't have to pay for the hosting. Please take into account that Netlify doesn't let you run server-side code (unless it's in a Netlify Function) but they do have support for certain frameworks like Next.js where they make this possible.
3. It needs to be secure.
4. It has to be as performant as possible. Studio is gonna grow into a big product that's gonna span multiple pages.
5. It would be great if it has a built-in way to create APIs. We're gonna need an API to power Studio and that will probably be exposed for anyone to use.
6. It should have good TypeScript support.
7. The `monaco-editor` package is gonna give us headaches, that's for sure. Make sure you can integrate it with the framework.

Lots of things to take into account when evaluating. An initial list of frameworks that come to my mind are:

1. Next.js
2. Vite
3. Remix
4. Gatsby

Learn more here: https://react.dev/learn/start-a-new-react-project#production-grade-react-frameworks.

## Rabbit holes

> Potential technical, design ... challenges

Well, we're gonna change the foundations of the project so there will be rabbit holes for sure. My advice is to try to keep it as simple as possible. This issue is just about migrating to a framework, not improving the codebase. Keep it simple. We can improve the codebase in other pull requests.

## Scope

> Describe a list of tasks or issues that needs to be done ( you don't need to have an exhaustive list of all the tasks in the beginning) 

- [x] Evaluate Next.js
- [x] Evaluate Vite
- [x] Evaluate Remix
- [x] Evaluate Gatsby
- [x] #684
- [x] #1063
- [ ] Add new tests with Cypress
- [ ] Add new DockerFile
- [ ] Implement code generation
- [ ] Document how to add new features and leverage React Server components
- [ ] Deployment in production (studio.asyncapi.com)

## Out of bounds

> What won't be part of the scope

[...]

PR #1062, opened 2024-04-03
feat: Use Next.js in studio-next

KhudaDad414:
**Description**
- this PR ports the studio from CRA to Next.js.
- it should compile and run in the dev environment.
- It should build without issues.

Some known issues and possible improvements:

- [ ] docker container file is missing in studio-next.
- [ ] tests are missing in studio-next after NEXT.JS migration.
- [ ] Code generation doesn't work in studio-next.
- [ ] Use CodeMirror instead of Monaco.

fixes #684

PR #1094, opened 2024-05-05
feat: add Cypress

aeworxet:
This PR introduces for `./apps/studio-next`:

- `Cypress` with testing minimal set of web application's UI elements due to upcoming migration to `./apps/design-system` (skeleton of the UI testing that will need to be reviewed after migration)
  - GUI testing is invoked with `npm run cy:open`
  - Headless end-to-end testing for different browsers is invoked with `npm run cy:e2e:*` (refer to the `cy:e2e:*` part of `package.json` for a list of available browsers (`edge` is reserved for Windows machines))

  image

- API route `http://localhost:3001/api/v1/generate` that only returns requests, just as a proof of concept that the server side is functioning (works only in a local dev environment; the deployed version is subject to CORS restrictions)

Deployment to Vercel is successful and can be viewed at https://asyncapi-studio-studio-next.vercel.app

Related to https://github.com/asyncapi/studio/issues/661
Related to https://github.com/asyncapi/studio/issues/1101


aeworxet:
@Amzani @KhudaDad414 
@asyncapi/bounty_team

aeworxet:
I started with the addition of `Cypress` so I could spot regressions during migration to `CodeMirror`.
[Comparison with PR #1062](https://github.com/KhudaDad414/studio/compare/use-nextjs...aeworxet:asyncapi-studio:feat-add-dockerfile-cypress-code-generation-codemirror)

@Amzani
A couple of questions:
- Is there a design file (Figma maybe?) where I can take unique names for all HTML elements from, so testing frameworks could find them in the webpage (same as I gave the name `data-testid="button-settings"` to the `Settings` button?)
- What names should I give to HTML elements' `data-*` attributes: `data-testid`, `data-test`, or `data-cy`?
- Should there be a separate spec file for components' testing, or, in the case of `Studio`, only full flow (e2e) makes sense?

Amzani:
@aeworxet 


> Is there a design file (Figma maybe?) where I can take unique names for all HTML elements from, so testing frameworks could find them in the webpage (same as I gave the name data-testid="button-settings" to the Settings button?)

Not for the existing studio, we only have one for the new improvements that will be integrated once we finish the NextJS migration: https://www.figma.com/proto/bPMB3lkMTOOMuKk0oGLuNm/Studio?type=design&node-id=96-2392&scaling=contain&page-id=0%3A1&starting-point-node-id=2%3A2

> What names should I give to HTML elements' data-* attributes: data-testid, data-test, or data-cy?

I vote for `data-test` for simplicity and in case we want to switch to an other testing tool in the future.

> Should there be a separate spec file for components' testing, or, in the case of Studio, only full flow (e2e) makes sense?

The goal is regression test, so I would suggest full flow (e2e), we can encourage components testing for new feature we add to the project. 

@KhudaDad414 WDY?

Amzani:
I would remove CodeMirror Integration from the scope of this issue for now, until we merge everything.

aeworxet:
I'm working on code generation for `Studio`, and after `POST` request to `https://api.asyncapi.com/v1/generate` I get
```
Version 3.0.0 is not supported.
Please use latest version of the specification.
```
Code, responsible for this is
https://github.com/asyncapi/parser-js/blob/master/src/ruleset/functions/isAsyncAPIDocument.ts#L25
https://github.com/asyncapi/parser-js/blob/master/src/constants.ts#L22
and the file which is searched for for `3.0.0` HAS it in `master`:
https://github.com/asyncapi/spec-json-schemas/blob/master/index.d.ts#L12

It needs to be checked on the [server](https://api.asyncapi.com) what version of `./node_modules/@asyncapi/specs/index.d.ts` it contains.
@smoya

Amzani:
@aeworxet we should remove the dependency with https://github.com/asyncapi/server-api in the NextJS version of studio as this repository will be deprecated soon.

So the idea is to directly use the [generator library](https://github.com/asyncapi/generator?tab=readme-ov-file#list-of-official-generator-templates).

Not all of them supports V3, AFAIK the following supports V3

- https://github.com/asyncapi/nodejs-template/
- https://github.com/asyncapi/html-template
- https://github.com/asyncapi/markdown-template

If < V3 all the generators should work as expected.

smoya:
More info about 

> So the idea is to directly use the [generator library](https://github.com/asyncapi/generator?tab=readme-ov-file#list-of-official-generator-templates).

@Amzani would you mind linking to where that decission has been made? Just to be in sync with https://github.com/asyncapi/server-api/issues/576

Amzani:
@smoya I was assuming that if only Studio uses this API there is no need to maintain it anymore thus deprecating the `server-api` repo and not migrate it to CLI to reduce complexity.

aeworxet:
> So the idea is to directly use the generator library.

@Amzani

@KhudaDad414 suggests the opposite, to stick with `server-api`.
So which one should it be?

Should I invest time into rewriting the logic for SSR, or should I just bugfix the current approach?

(https://github.com/asyncapi/studio/pull/1094#issuecomment-2118721184 is still relevant then)

KhudaDad414:
@aeworxet IMHO it makes sense to bug fix the current code base. By that I mean sticking to the server-api approach.

As Fran said in the original issue, migration is already hard enough, trying to implement new features at the same time would quadruple the complexity.

aeworxet:
@KhudaDad414 
'Code Generation' functionality still requires checking of the `./node_modules/@asyncapi/specs/index.d.ts` version on `api.asyncapi.com` as per https://github.com/asyncapi/studio/pull/1094#issuecomment-2118721184 since it is the server that is returning the error.

image

KhudaDad414:
@aeworxet Since you get the error on the [current Studio](studio.asyncapi.com) as well, I would consider resolving it to be out of scope for this Bounty Issue.

We can have another issue and discuss it over there.

aeworxet:
@KhudaDad414

No, this is the one that is being migrated

`studio/fork/apps/studio-next$ pnpm run dev`

and this server's issue is directly related to migration.

aeworxet:
I'm investigating options for deployment of `Studio` to a Docker container, and the possibility that the server part [might become completely independent](https://github.com/asyncapi/server-api/issues/485), up to being hosted fully in a CLI with invocation through `asyncapi server start`, led me to a question if deployment of Studio to a Docker container as a full Next.js app (with API Routes, SSR, and others) is needed at all.

The most runtime-dynamic part of Studio currently is the generation of templates, and it can be done through a POST request to an API from a static website (it is made this way currently, in fact.)

I'm not sure [OG preview generation](https://github.com/asyncapi/studio/issues/224) is a mandatory functionality for a containerized Studio, it can be left reserved for https://studio.asyncapi.com

I have deployed to https://asyncapi-studio-studio-next.vercel.app a build of `Studio` with `output: 'export'` ([explanation what 'export' build is](https://stackoverflow.com/a/61835649), [features that are NOT supported](https://nextjs.org/docs/pages/building-your-application/deploying/static-exports#unsupported-features).)

The whole 'export' build is nine (9) Mb, it has First Contentful Paint `0.4 s` (+ better overall performance score in Lighthouse,) and it doesn't seem to lack any functionality.

Should this version of Studio's build be used for a Docker container?

Amzani:
> The most runtime-dynamic part of Studio currently is the generation of templates, and it can be done through a POST request to an API from a static website (it is made this way currently, in fact.)

We want to enable long term features like authentication for instance. generation of templates using API to POST should be something we could migrate IMO as we don't want to maintain an entire repository just for 1 single `POST /generate` method instead of using directly the `generator` library.

> Should this version of Studio's build be used for a Docker container?

How complex is having everything in Docker image?

aeworxet:
@Amzani

> How complex is having everything in Docker image?

It is [supposed to be straightforward](https://github.com/vercel/next.js/tree/canary/examples/with-docker#using-docker) but I ran into an error on a command that executes fine on desktop, so I'm investigating it.

image

aeworxet:
I hoped to have for `ouput: 'export'` a 20 MB Docker image (5 MB alpine linux + 9 MB static `Studio` + several MB for nginx,) but for `output: 'standalone'` Node.js v18 alone takes 125 MB, and overall Docker image size is 282 MB (compiled `Studio` is 20 MB, though.) The previous Docker image was 91 MB.

A request to the API works as a POC.

image


I attempted to delete something unneeded, but it seems that everything in this Docker image is used for one process or another:

image


Try image:

```bash
cd ./apps/studio-next
docker build .
docker images # the freshly built Docker image is most probably the topmost one
docker run -p 3001:3001 [image_name]

http://localhost:3001
```

apps/studio-next/Dockerfile
KhudaDad414, line 1: IMHO it is better to follow https://turbo.build/repo/docs/guides/tools/docker and create a docker image that way. it takes forever to install the dependencies (more than 30 minutes). I think it's because we are not leveraging `turbo prune` to get rid of the unwanted packages.

KhudaDad414: Another point: It takes forever to run post-install scripts of Cypress (I think it installs all of the browsers with each install, like Puppeteer). there should be a way to disable that when building a container. something like `--ignore-scripts` in npm. 🤔

apps/studio-next/package.json
```diff
@@ -6,7 +6,14 @@
     "dev": "next dev -p 3001",
     "build": "next build",
     "start": "next start",
-    "lint": "echo 'No linting configured'"
+    "lint": "next lint",
+    "docker:build": "docker build -t asyncapi/studio:latest .",
```
KhudaDad414, line 10: docker build should be done at the root of the project (so turborepo is able to prune the unwanted packages). maybe something like: `"build:next-studio:docker": "docker build -t asyncapi/studio:latest -f StudioNextDockerfile ."` at the root.

apps/studio-next/src/app/api/v1/generate/route.tsx
```diff
@@ -0,0 +1,6 @@
+import { NextResponse } from 'next/server';
```
KhudaDad414, line 1: Not sure if we can use [Server Actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) instead 🤔 ?

KhudaDad414 (changes requested):

apps/studio-next/package.json
aeworxet: Simply making sure that Turborepo will not start to be turned into a monolith again after this change.

Might it be better, just in case, to keep possibility to dockerize each app separately with [Docker Compose](https://docs.docker.com/compose/intro/features-uses), like it's done in https://github.com/dhaaana/turborepo-with-docker:

- https://github.com/dhaaana/turborepo-with-docker/blob/main/docker-compose.yml
- https://github.com/dhaaana/turborepo-with-docker/blob/main/apps/api/Dockerfile
- https://github.com/dhaaana/turborepo-with-docker/blob/main/apps/web/Dockerfile

?
(meaning `Dockerfile` will stay where it is now, but `Docker Compose` setup will be added)

KhudaDad414: If it's possible to [publish the separate docker image](https://github.com/asyncapi/studio/blob/master/.github/workflows/publish-docker-image.yml) after this, I have no issue with it.

apps/studio-next/src/app/api/v1/generate/route.tsx
aeworxet: @jerensl

I searched the Internet back and forth and couldn't find answers to several simple questions.
You [seem to be](https://github.com/asyncapi/studio/issues/1118) very knowledgeable in Next.js Server's matters, so I would like to consult with you on this topic.

Server Actions are meant for performing data fetching and mutations on the Next.js Server.
To my understanding, this implies that Server ALREADY exists as an independent entity, in order for a Server Action to have an environment to be executed in.
In this particular case, right now, full Server consists of one API route.
Thus, if this one and only API route is replaced by a Server Action, the Server will simply vanish, and Server Action will have no place to be executed in.
Besides, this API route is responsible for Server Response, which requires exactly the `next/server` package, and not for data fetching from a third-party resource or a mutation.

Taking this into account, does Server Action even make sense in this case, or should the structure with one API route remain (for now, the final architecture of the `Studio`'s Server-Side is not defined yet, and this route is added only as a Proof Of Concept and a stub in place where the future Server will be?)

Please correct me if I'm getting something wrong in the new for me concept of Server Actions.

jerensl: @aeworxet 

> I searched the Internet back and forth and couldn't find answers to several simple questions. You [seem to be](https://github.com/asyncapi/studio/issues/1118) very knowledgeable in Next.js Server's matters, so I would like to consult with you on this topic.

I still new to this Reace Server Components (RSC), so I will answers as much as I know
 
> Server Actions are meant for performing data fetching and mutations on the Next.js Server. To my understanding, this implies that Server ALREADY exists as an independent entity, in order for a Server Action to have an environment to be executed in. In this particular case, right now, full Server consists of one API route. Thus, if this one and only API route is replaced by a Server Action, the Server will simply vanish, and Server Action will have no place to be executed in. Besides, this API route is responsible for Server Response, which requires exactly the `next/server` package, and not for data fetching from a third-party resource or a mutation.
> 
> Taking this into account, does Server Action even make sense in this case, or should the structure with one API route remain (for now, the final architecture of the `Studio`'s Server-Side is not defined yet, and this route is added only as a Proof Of Concept and a stub in place where the future Server will be?)

The thing about [Api Route](https://nextjs.org/docs/pages/building-your-application/routing/api-routes) which their rebranding as [Route Handler](https://nextjs.org/docs/app/building-your-application/routing/route-handlers) is not supposed to be replaced by Server Actions but different type options in the Architecture which would be considered during the design

https://www.youtube.com/watch?v=PtX5PF17tlQ

> Please correct me if I'm getting something wrong in the new for me concept of Server Actions.

Server Actions are supposed to work on top of the React Server Component(RSC) but the majority of the applications inside [StudioWrapper](https://github.com/asyncapi/studio/blob/master/apps/studio-next/src/components/StudioWrapper.tsx) do not leverage RSC even for NextJS features like SSR/SSG do not work there

Static Side Generator -> render JSX to HTML during built time
Server Side Rendering -> render JSX to HTML at server when requested by a Client (give what u need only)

This one of the good resources Dan Abramov explains what are the responsibilities of SSR(What NextJS did) and RSC, and how they work together interchangeably
https://github.com/reactwg/server-components/discussions/5

aeworxet: @KhudaDad414 
Would it be safe to assume that Server Actions should not be used currently?

apps/studio-next/package.json
aeworxet: `turbo prune studio-next --docker` copies to `./out`'s `node_modules` containing `pnpm`'s symlinks, which, after being copied, immediately become invalid (because they point to previous relative destinations,) but doing `pnpm install` each time there's a need to build a Docker image is not an option either. So I'm trying different configurations to achieve local caching in at least some form, and each build takes time. Thus delay.

aeworxet: I changed approach in favor of pushing into the Docker image the **already compiled** `Studio`.
I took advantage of `turbo build` cache, and now if there is a fresh `turbo`'s build cache, then the generation of the Docker image takes not even seconds but milliseconds.

image

Please try running
```
turbo build
turbo run build:studio-next:docker
```
in the root of the repo and check.

The generated Docker image (became 294MB) is run with
```
docker run -p 3001:3001 asyncapi/studio
```

KhudaDad414: > turbo prune studio-next --docker copies to ./out's node_modules containing pnpm's symlinks, which, after being copied, immediately become invalid

That's why we use ` turbo prune` to get rid of the unwanted packages before running `pnpm install`. after we have made sure to separate the wanted from the unwanted packages.

after that we install the packages inside of the docker container.

> I changed approach in favor of pushing into the Docker image the already compiled Studio.

well, we will run the docker build command on the GitHub Actions workflow. so there will not be any *already compiled* studio to build the docker image from.

We would still have to install dependencies for the entire mono-repo, re-build the entire project, then use the built project to build the docker image.

even if it's easier for us to build, think about the user of the image. if we push one big un-cacheable layer (like we currently do) every time something changes (even if it doesn't affect studio-next), it will make the build command for them to take soooo much longer.

why not follow the [official approach](https://turbo.build/repo/docs/guides/tools/docker#example) which would build the studio on layers and then below layers will be cached anyway for the end user, so they will only download what they need( for example if no new dependency added, they will not install the dependencies again, they will get the source and continue from there).

At the end, it doesn't matter what the final size of the image is, docker won't install all of it, it will cache the steps that are possible to cache and only builds from the layer that the change is detected. we can help docker by separating our steps(first, copy package.json, then install dependencies, then copy the source code, then build, etc...) so it can properly cache each step.

apps/studio-next/src/app/api/v1/generate/route.tsx
KhudaDad414: @aeworxet you are right, we first need to decide how we are going to architecture the component for the `app` directory.

apps/studio-next/package.json
KhudaDad414: btw, I am not an expert in docker. please let me know if I misunderstand something here. 😆

aeworxet: `turbo prune studio-next --docker` creates in `./out/full/apps/studio-next/node_modules` set of symlinks to all used Node.js' modules, which are invalid even on local filesystem, let alone in the Docker image filesystem after they are pushed to it.

image
image

Is it different for you?

To my understanding, `turbo prune studio-next --docker` would be useful if `Studio` used `npm`, but in the case of `pnpm`, it's useless exactly due to symlinking of Node.js modules. That's why, after trying to get it working for a week, at the end of the day, I changed approach to pushing to the Docker image the already locally built binaries.

Might I simply be doing something wrong?


```
"build:studio-next:docker": {
  "dependsOn": ["^build:studio-next:docker"]
}
```
is hardcoded to run `next build` as the first step
https://github.com/asyncapi/studio/pull/1094/files#diff-c493037e1471391d87ade822ee79d21c8ba48c14063c5feafdf629443b19cc0eR10

So either there is already a fresh build and the `turbo`'s cache will be utilized, or `Studio` will be built from scratch anyway.

Aside from that, I also think that release process can have two simultaneous deployments:
- Build and deploy to the **studio.asyncapi.com**.
- Reuse already built binaries to generate a Docker image and deploy it to the **Docker hub**.

Such a release process will save time and space, and it seems quite logical overall.

KhudaDad414: @aeworxet 
Why are we copying the `node_modules` folder to the docker?
I think we are supposed to ignore it and install it fresh in the docker environment as a separate step.
This way it will be cached by docker and we never have to reinstall the packages unless a package.json file changes.

can you try adding `node_modules` to the `.dockerignore` file?

aeworxet:
@KhudaDad414
How should I handle these ESLint's issues? They are not mine, but should I rewrite them anyway, disable ESLint rules, or should these simply be ignored due to future rewrite?
```
./src/components/Navigation.tsx
88:17  Warning: Refactor this function to reduce its Cognitive Complexity from 16 to the 15 allowed.  sonarjs/cognitive-complexity

./src/components/Navigationv3.tsx
159:17  Warning: Refactor this function to reduce its Cognitive Complexity from 18 to the 15 allowed.  sonarjs/cognitive-complexity

./src/components/common/Switch.tsx
14:27  Error: 'e' is defined but never used.  @typescript-eslint/no-unused-vars
```

KhudaDad414:
@aeworxet they should be ignored.
we are already ignoring them in sonar.
https://github.com/asyncapi/studio/blob/master/.sonarcloud.properties

apps/studio-next/package.json
KhudaDad414: I put together this one which I think you can use to improve upon:
btw you need to enable docker buildKit for pnpm caching to work. not sure how we can enable it in Github Actions.
```dockerfile
FROM node:18-alpine AS base


FROM base AS builder
RUN apk update
RUN apk add --no-cache libc6-compat

ENV PNPM_HOME="/pnpm"
ENV PATH="$PNPM_HOME:$PATH"
RUN corepack enable

WORKDIR /app
RUN pnpm add -g turbo@^1

COPY . .
ENV NODE_ENV production
RUN turbo prune studio-next --docker

FROM base AS installer
ENV DOCKER_BUILDKIT=1
RUN apk update
RUN apk add --no-cache libc6-compat

ENV PNPM_HOME="/pnpm"
ENV PATH="$PNPM_HOME:$PATH"
RUN corepack enable
WORKDIR /app

COPY .gitignore .gitignore
COPY --from=builder /app/out/json/ .
COPY --from=builder /app/out/pnpm-lock.yaml ./pnpm-lock.yaml

RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --ignore-scripts --frozen-lockfile

RUN pnpm run build:studio-next

FROM base AS runner

RUN addgroup --system --gid 1001 nodejs
RUN adduser --system --uid 1001 nextjs
USER nextjs

COPY --from=installer /app/apps/studio-next/next.config.js .
COPY --from=installer /app/apps/studio-next/package.json .

COPY --from=installer --chown=nextjs:nodejs /app/apps/studio-next/.next/standalone ./
COPY --from=installer --chown=nextjs:nodejs /app/apps/studio-next/.next/static ./apps/studio-next/.next/static
COPY --from=installer --chown=nextjs:nodejs /app/apps/studio-next/public ./apps/studio-next/public


CMD node apps/studio-next/server.js

```

aeworxet: `node_modules` [were ignored initially](https://github.com/asyncapi/studio/pull/1094/files#diff-2f754321d62f08ba8392b9b168b83e24ea2852bb5d815d63e767f6c3d23c6ac5L3).
`pnpm` was installing them from scratch inside the Docker image.
[You said](https://github.com/asyncapi/studio/pull/1094#discussion_r1648838607) Installing them takes too long.
I changed approach to pushing to the Docker image the pre-compiled as `standalone` Next.js app.
`node_modules` are an integral part of the `standalone` version (`export` version is fully static.)
Docker didn't copy `node_modules` to the image because they were in `.dockerignore`, making `standalone` Next.js app unusable.
I had to remove the `node_modules` from `.dockerignore`.

It is already optimized during compilation of the Next.js app for production, though, and takes only 152 MB of code, not 1.9 G of root `node_modules`.

aeworxet: @derberg
Can you please suggest a Docker expert?
I'm afraid we're starting to go in circles here.

KhudaDad414: @aeworxet 

> I had to remove the node_modules from .dockerignore.

I never suggested that you remove the `node_modules` from docker container or change your approach. all I said was to add the `--ignore-scripts` flag so the browser is not downloaded by Cypress during installation.

> You said Installing them takes too long.

Yes, it takes too long. but one part of the solution is to not run the post-install scripts which [I suggested here](https://github.com/asyncapi/studio/pull/1094#discussion_r1648838607) and the second part would be to make the pnpm use it's cached packages which I [suggested here](https://github.com/asyncapi/studio/pull/1094#discussion_r1668359299).

> I changed approach to pushing to the Docker image the pre-compiled as standalone Next.js app.

Well, the problem with that approach is that docker can't cache the result of the build steps and the user of the docker image has to pull the entire image for each update. which [I pointed out here](https://github.com/asyncapi/studio/pull/1094#discussion_r1668291253).

> It is already optimized during compilation of the Next.js app for production, though, and takes only 152 MB of code, not 1.9 G of root node_modules.

What do you think the image size would be if you follow [the official approach](https://turbo.build/repo/docs/guides/tools/docker)? Which I suggested twice ([here](https://github.com/asyncapi/studio/pull/1094#discussion_r1668359299), and [here](https://github.com/asyncapi/studio/pull/1094#discussion_r1648834593)) by now in this thread.

Let's take a look at the sizes with the two approaches:
Screenshot 2024-07-08 at 14 05 36
As you can see the official approach's image size is even smaller than what [you claim](https://github.com/asyncapi/studio/pull/1094#discussion_r1662117391) your approaches' size is.
(Maybe something is wrong here since your image size comes up as 3.19GB on my machine.)

> I'm afraid we're starting to go in circles here.

I don't exactly get what you mean.
1) You implemented a solution
2) the solution had a problem(pnpm packages took too long to install).
3) I pointed out the problem and suggested [a possible solution](https://github.com/asyncapi/studio/pull/1094#discussion_r1648838607)
4) you changed the ENTIRE approach.
5) I pointed out the [problems with the new approach](https://github.com/asyncapi/studio/pull/1094#discussion_r1668291253) and suggested [a solution](https://github.com/asyncapi/studio/pull/1094#discussion_r1668359299) which resolves the problem. (by not running post-install scripts and using pnpm's cache)
6) you claim we are going full circle and need someone else to review your work?

aeworxet: I clearly [misperformed](https://github.com/asyncapi/studio/issues/661#issuecomment-2215396594) on this Bounty Issue.
Should Docker-related code be removed so other changes could be merged?

aeworxet: Just built the Docker image with `turbo build:studio-next:docker`

image

The difference is strange.

aeworxet: Maybe you should clone this branch from scratch and do in the repo's root
```bash
pnpm install
docker builder prune -a    # <-- remove Docker's build cache
turbo build:studio-next:docker
```
?

Just built again with these steps

image

aeworxet:
Disabled ESLint's rules on warnings and the error.

aeworxet:
SonarCloud is frozen.

image

Sergio was dealing with this issue several days ago
https://github.com/asyncapi/parser-js/pull/1008#issuecomment-2208211553

apps/studio-next/package.json
KhudaDad414: Upon some reflection I think my tone might have come across as too critical and for that I apologize.

My intentions were to provide technical feedback not to criticize you or your efforts personally.

I know you have worked hard on the issue and I don't think it was just for you to be suspended in the bounty program or not getting the reward for this bounty issue.

If you think it's best to remove the docker and merge the remaining code, let's do that.

KhudaDad414: > Maybe you should clone this branch from scratch and do in the repo's root
> ```bash
> pnpm install
> docker builder prune -a    # <-- remove Docker's build cache
> turbo build:studio-next:docker
> ```
> ?
> 
> Just built again with these steps
> 
> image
> 

I cleaned the project reinstalled the packages and built the image with `turbo build:studio-next:docker`

The size now shows up correctly.

aeworxet: What do you think about reusing the same compiled code for two simultaneous deployments now?

> release process can have two simultaneous deployments:
> 
> - Build and deploy to the **studio.asyncapi.com**.
> - Reuse already built binaries (with `turbo` cache) to generate (push them into) a Docker image and deploy it to the **Docker hub**.

This process will guarantee that there is one same `Studio`, built from one same commit, with the same hashes, in both places.
(I already have a situation right now when the same bug gets reproduced in one place, doesn't in another, and gives a third error in the local dev)

KhudaDad414: @aeworxet

### Building in One Environment and Running in Another Is Generally a Bad Idea
As you know, we are using linux to run random workflows. However, when it comes to publishing or testing, we are using a matrix of (Linux, Windows, and Mac) to [build, publish or test](https://github.com/asyncapi/cli/blob/master/.github/workflows/if-nodejs-release.yml) our code.

Why? Because we want to make sure that it is built within the environment that it is going to run on. How can you be sure that if you build something on Linux, it will run on Windows? You might argue that they are different since docker has it's own Engine on top of the OS and it will work. You might be right, but we need to test that theory. Is it worth it to go though the trouble, I don't think so.
All I know is that some packages are platform-dependent. By installing in the Docker image itself, we make sure that all our packages are installed, built, and run in the same environment.

That's why building in the Docker image itself is always better.

### Transparency
When I look at the Dockerfile, I should be able to see that yes, I am actually downloading the source code and building it myself on my machine. This gives me peace of mind that nobody is pushing some fishy code into my environment. Most of the ecosystem now doesn't trust pre-built packages because of incidents like the [XZ fiasco](https://www.youtube.com/watch?v=bS9em7Bg0iU).

### Docker Cache
As I mentioned previously, Docker is really smart when it comes to building and running your image. If you follow best practices, it can help your package be as slim as possible.

For example, if you haven't changed your package.json or lock file, why would you want to install your packages from scratch? Well, you don't want that. You want to get the source code, which is probably a few MB, use the cached version of the package installation layer, and build on top of that, right? Well, Docker is going to do exactly that. But if you are pushing an already built 290MB of code in a single layer, Docker won't be able to do anything. Even if the next version only changes by one byte, it has to invalidate that layer and install the entire 290MB of code again.

So in a Dockerfile, it is always better to break down your build, installation, and run layers as much as you can. More layers mean Docker is able to cache more efficiently when `building` and when `running`.

### Harder to Maintain
It is hard to maintain a separate and custom workflow to do things. People expect simplicity. If I want to build an image, I would naturally expect the `docker build` command to work. why does it have to be dependent on other processes like installation and building?

If you build one more layer of abstraction on top and use a script, it is suspicious to say the least. Contributors/Users will have to check what exactly you are doing in that script and why they can't just build it with a simple `docker build` command.

aeworxet: Thanks for investing time in explaining all that.
I now understand that I should have taken specific end users' needs into account also, not only aim at easing the release process for the vendor (AsyncAPI.) 
The Docker build should definitely be developed by someone who understands Docker better.
Removed Docker-related code and changed the PR's description accordingly.

aeworxet: Thanks for investing time in explaining all that.
I now understand that I should have taken specific end users' needs into account also, not only aim at easing the release process for the 'vendor.' 
The Docker build should definitely be developed by someone who understands Docker better.
Removed Docker-related code and changed the PR's description accordingly.

aeworxet:
Can this PR be merged, or is there anything else that needs to be done on it?

Amzani:
@aeworxet could you fix the conflict ?
Thanks

aeworxet:
@Amzani
Fixed.

aeworxet:
I had to rename

`ConvertVersion` -> `AsyncAPIConvertVersion`

which was [renamed recently](https://github.com/asyncapi/converter-js/pull/269/files#diff-2ee8da4ec0f3c043c7e097851af07383ae3ce13022973c8727312e06bf2b89b3R6) in `@asyncapi/converter` [v1.5.0](https://github.com/asyncapi/converter-js/issues/233#issuecomment-2285378563).

In the change's PR only the name was changed, so all functionality is supposed to remain the same, but pinging @jonaslagoni just in case.

Amzani (approved):
LGTM

aeworxet:
@KhudaDad414
Can you please dismiss your change request since I have [removed](https://github.com/asyncapi/studio/pull/1094#discussion_r1671804394) all Docker-related code?

KhudaDad414 (approved):

Issue #1105, opened 2024-05-29
Improve JSON Schema Editor


KhudaDad414:
### Problem

The work started in #1065 is a great start, but in my opinion, a functional visual JSON Schema editor needs more attention. There are some missing features I think are essential for it to be considered production-ready.

### Solution

First of all, I propose creating a separate package for the visual JSON Schema editor. This will help us to publish it separately and create a separation of concerns in the studio as well.

### Rabbit holes

...

### Scope

- [ ] Move the JSON Schema editor to a separate package.
- [ ] Improve the JSON Schema editor's "Add New Property" feature.
- [ ] Make the JSON Schema editor field names editable.
- [ ] Make the JSON Schema editor's look and feel follow the provided design.
- [ ] Support combiners (anyOf, oneOf, allOf) in the JSON Schema editor.
- [ ] Design and implement a basic type selector for the JSON Schema editor.

### Out of bounds

Supporting $refs is out of scope.
A complete type editor would be out of scope. (too complex and should be considered in a separate pitch after the design is finalized)

### Success criteria

- Have a published package called `@asyncapi/json-schema-editor` that contains a React component for editing JSON Schemas.
- Make adding a new field easier. When I click on "New Property", a new empty property should appear with the default type of `string`.
- Make the property names editable.
- Make the JSON Schema editor look and behave like the provided design.
- Support `anyOf`, `oneOf`, and `allOf` combiners in objects as well as in arrays.

KhudaDad414:
/progress 50

DropdownMenu has been refactored to be more flexible to be used in JSON Schema Type Editor. #1108

PR #1115, opened 2024-06-11
Add DropdownChipInput

KhudaDad414:
**Description**

- the new component is just like ChipInput. but the set of options has to be defined.
- It's going to be used for #1105


princerajpoot20 (approved):
LGTM

PR #1121, opened 2024-07-05
fix(cold-start): add another component as layer for lazy loading

### LINKED ISSUE #1118: Response time for NextJS version seems too high
## Description

I'm working with @helios2003 in https://github.com/asyncapi/studio/issues/224. In particular, I'm its mentor through GSoC2024.
As part of such project, we evaluated the possibility to run the AsyncAPI Parser-JS and parse AsyncaPI documents loaded via the `base64` query param. 
@helios2003 made [a test measuring response time ](https://github.com/asyncapi/studio/issues/224#issuecomment-2160148396)with and without the addition of the parsing (which would made the page from being statical to ssr). Our surprise came when we realized the NextJS version hosted in https://github.com/asyncapi/studio/issues/224 took ~4 seconds to just serve the HTML for the Studio page (without even loading any doc on it!), just plain `/`.

Today I decided to run another test, and confirmed the findings. However, a weird caching mechanism is happening. 

Let me share with you 3 consecutive requests I made and the results:

| Request # | Cache-Status                           | Cache-Status meaning                                                             | Seconds total |
|-----------|----------------------------------------|----------------------------------------------------------------------------------|---------------|
| 1st       | `[Next.js; hit, Netlify Edge; fwd=miss]` | Missed Cache at Netlify level, Hit Cache of NextJS                               | 4.072090      |
| 2nd       | `[Netlify Edge; hit]`                   | Hi Cache at Netlify level, meaning response is served from Netlify edge directly | 0.123647      |
| 3rd       | `[Next.js; hit, Netlify Edge; fwd=miss]` | Missed Cache at Netlify level, Hit Cache of NextJS                               | 0.572575      |

My assumption and understanding is that, for some reason: 

- During the first request there is stuff being rendered at server level dynamically (runtime) instead of something precompiled at build time.
- The second request does a hit in Netlify Edge cache, the reason why it takes an insignificant time to give a response since is not even reaching Studio app.
- The third request hits another Netlify edge server which still has no cached content (normal), then the request reaches the NextJS app and, in this case, the content is finally served by the NextJS cache.

The point is that I don't expect the first call to take 4 seconds but just as the third request since a request made to the root page should always give the same response (static). Besides that, no clue why the `cache-status` header in the 1st request saying the content was served as cached from NextJS.

cc @Amzani @helios2003 @KhudaDad414

jerensl:
Using `next/dynamic` in server components that loading a client component is still including everything in server side rendering, and that client component doesn't know it needs to be split into another chunk in browser chunks.

Related: https://github.com/vercel/next.js/issues/49454#issuecomment-1830053413

**Description**
I set up my own deployment here https://jolly-toffee-18b732.netlify.app and tested the results below
Screenshot 2024-07-05 145715

**Related issue(s)**
Fixes #1118

### DIFF
--- a/apps/studio-next/src/components/StudioEditor.tsx
+++ b/apps/studio-next/src/components/StudioEditor.tsx
@@ -1,0 +1,10 @@
+'use client'
+import dynamic from 'next/dynamic'
+const StudioWrapper = dynamic(() => import('@/components/StudioWrapper'), {ssr: false})
+
+
+export default async function StudioEditor() {
+  return (
+    <StudioWrapper />
+  )
+}
--- a/apps/studio-next/src/app/page.tsx
+++ b/apps/studio-next/src/app/page.tsx
@@ -1,7 +1,6 @@
-import dynamic from 'next/dynamic';
-const StudioWrapper = dynamic(() => import('@/components/StudioWrapper'), {ssr: false})
 import { Metadata } from 'next';
 import ogImage from '@/img/meta-studio-og-image.jpeg';
+import StudioEditor from '@/components/StudioEditor';
 
 export const metadata: Metadata = {
   metadataBase: new URL('https://studio-next.netlify.app'),
@@ -25,6 +24,6 @@ export const metadata: Metadata = {
 }
 export default async function Home() {
   return (
-    <StudioWrapper />
+    <StudioEditor />
   )
 }


apps/studio-next/src/components/StudioEditor.tsx
KhudaDad414, line 1: Great work @jerensl. I tested the result and it seems like it is working. the only issue that I see here is the layers that we are adding. StudioEditor -> StudioWrapper -> CodeEditor. do you think we can do something about it? maybe combine them into one wrapper? or at least two?

if that's not possible, can you also add a comment on why this wrapper has been created?

KhudaDad414 (commented):
Great work @jerensl 👏 .

apps/studio-next/src/components/StudioEditor.tsx
jerensl: The reason we added `StudioEditor` is that we are unable to make root-level pages(/) as client components, so we added an extra component as a client component to call the StudioWrapper as code splitting

KhudaDad414 (changes requested):
one last thing. can you also add a changeset so it trigger the release process?

KhudaDad414:
@jerensl you don't need to version the PR. all that is required is having a changeset file. the CI/CD pipeline will take care of versioning.

you can do that by running `pnpm run changeset` follow the instructions and the changeset file will be created. after that push the changeset file. that's it. you don't need to run `pnpm run version-packages`. it will be done by the pipeline.

jerensl:
> @jerensl you don't need to version the PR. all that is required is having a changeset file. the CI/CD pipeline will take care of versioning.
> 
> you can do that by running `pnpm run changeset` follow the instructions and the changeset file will be created. after that push the changeset file. that's it. you don't need to run `pnpm run version-packages`. it will be done by the pipeline.

Thanks, I made the changes in the latest commit

KhudaDad414 (approved):
LGTM

jerensl:
## Updated

In NextJS 15, the [cache is no longer default](https://nextjs.org/blog/next-15-rc#caching-updates). I'm also testing it on my deployment https://jolly-toffee-18b732.netlify.app/, and the cache lives longer.
Screenshot 2024-08-02 185501

I assume that NextJS tries to invalidate the cache when it is already being [cached by Netlify](https://docs.netlify.com/platform/caching/#default-caching-behavior) and did the same cache as well this led to a cold start because it stale while revalidated always looking at the first [cache-store](https://developer.chrome.com/docs/workbox/caching-strategies-overview/#stale-while-revalidate).

So I will mark this as partially fixed, the rest will be done by updating to NextJS 15 when it's released, from now it's still on the release candidate

smoya:
Can Studio owners prioritize this issue? 

cc @magicmatatjahu @KhudaDad414 @Amzani

Amzani (approved):
LGTM

Shurtu-gal:
@smoya @jerensl where was this left. I am seeing the approvals were already there.

PR #1106, opened 2024-05-31
feat: add og tags to studio-next

helios2003:
**Description**
Adds static open graph tags to the studio-next application. 

**Screenshots**
For Twitter:
Screenshot 2024-05-28 234008

For Whatsapp
Screenshot 2024-05-28 234650

Network Tab
Screenshot 2024-05-28 234752

**Known limitation(s)**
The AsyncAPI image does not get rendered when sharing the link on Slack but works on all other social media platforms.

**Related issue(s)**

Since https://github.com/asyncapi/studio/issues/224 has been chosen for this year's GSoC I will add incremental improvements and bring it closer to the actual project which includes making the ``og-tags`` as well the `og-image` dynamic.

PR #1122, opened 2024-07-09
feat: make the open graph tags dynamic based on the ``base64`` and ``url`` query parameters

### LINKED ISSUE #1118: Response time for NextJS version seems too high

helios2003:
This PR addresses the issue mentioned in [#224](https://github.com/asyncapi/studio/issues/224). I have added the following features in this PR.

- The studio version I am testing on is hosted here: [https://studio-studio-next.vercel.app](https://studio-studio-next.vercel.app). (I will switch this to Netlify deployment once we are able to resolve [#1118](https://github.com/asyncapi/studio/issues/1118))
- The Image Generator service in charge of creating the dynamic image is hosted here: [https://ogp-studio.netlify.app](https://ogp-studio.netlify.app)
- The code for the image Generator service can be found here: https://github.com/helios2003/ogp-studio

## Features Added

- [x] Add a middleware that captures Open Graph crawler requests when they hit the website and rewrites the URL to the route `api/crawler`. This is done to serve minimal HTML when the crawler hits the page, reducing the chance of timeout.
- [x] Parse the URL and send it to the Image Generator service to add the dynamic image.
- [x] Show the default `og:image` if there is an invalid base64 string provided.
- [x] Fall back on the usual behavior added in [#1106](https://github.com/asyncapi/studio/pull/1106) if there isn't a base64 string.

## Todos

- [ ] Address unusual behavior observed like Slack showing the default image even when the document is valid.
  - For example: if we go to [https://studio-studio-next.vercel.app](https://studio-studio-next.vercel.app) and use MQTT template, then the default behavior is observed instead of dynamic tags.
  - Sometimes, Slack crawlers arent able to detect the open graph image.
- [ ] Figuring out why sometimes the dynamic `og:image` does not appear on site like Whatsapp.
- [ ] Add additonal crawlers to the list.

cc : @smoya


apps/studio-next/src/middleware.ts
KhudaDad414, line 1: The configuration and this middleware achieve the purpose, right? I don't understand why we are adding both.

KhudaDad414 (changes requested):
left some comments.

helios2003:
Thank you for the review @KhudaDad414. I will add the suggested changes.

helios2003:
Also, I had a few questions:
- Is it okay if I hardcode the popular social media site crawlers as I have done now or is it better if we use a library like https://github.com/monperrus/crawler-user-agents as suggested to me by @smoya.  The main problem I could think of is that by including all the potential sets of crawlers as mentioned in the repository we could be causing rewrites for nonopen graph crawlers as well which is not required. But @smoya mentioned that he does not expect other bots to use the query parameters like that of open graph crawlers.
- Also, could you please let me know your thoughts on https://github.com/helios2003/ogp-studio and its code?

apps/studio-next/src/helpers/parser.ts
```diff
@@ -0,0 +1,40 @@
+import { Parser } from '@asyncapi/parser';
+import { DocumentInfo } from '@/types';
+
+export default async function parseURL(base64Document: string): Promise<DocumentInfo | null> {
```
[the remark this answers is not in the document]
helios2003: So do you recommend keeping the `studio` version, or using this code?

apps/studio-next/next.config.js
```diff
@@ -36,6 +36,20 @@ const nextConfig = {
 
     return config
   },
-}
-
-module.exports = nextConfig
+  async rewrites() {
+    return [
+      {
+        source: '/:base64',
```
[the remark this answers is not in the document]
smoya: Yes, we should support both `base64` and `url` use cases as stated in https://github.com/asyncapi/studio/issues/224

smoya:
> * Is it okay if I hardcode the popular social media site crawlers as I have done now or is it better if we use a library like https://github.com/monperrus/crawler-user-agents as suggested to me by @smoya.  The main problem I could think of is that by including all the potential sets of crawlers as mentioned in the repository we could be causing rewrites for nonopen graph crawlers as well which is not required. 

If you don't use a complete list of crawlers, you will end up facing issues like the one you mentioned:

> Figuring out why sometimes the dynamic og:image does not appear on site like Whatsapp.

This is happening most probably because you are not considering [Whatsapp User-Agent(s)](https://github.com/monperrus/crawler-user-agents/blob/master/crawler-user-agents.json#L2141).

> But @smoya mentioned that he does not expect other bots to use the query parameters like that of open graph crawlers.

AFAIK, indexer robots such as Google **only** follow links, meaning that unless there is a published link somewhere to a link such as `https://studio.asyncapi.com/?base64=<base64_doc>` or the variant with `url` query param, the parsing won't happen. In fact, it does make sense to do the parsing and return the proper title, description, etc to the results of such indexer bot request so it appears in the results (like google results).

smoya:
> * Also, could you please let me know your thoughts on https://github.com/helios2003/ogp-studio and its code?

Will that work in a Netlify setup? I mentioned an alternative in case it doesn't in https://github.com/asyncapi/studio/issues/224.

apps/studio-next/next.config.js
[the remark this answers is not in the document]
helios2003: Yes sure, I will add it.

helios2003:
> Will that work in a Netlify setup? I mentioned an alternative in case it doesn't in #224.


Yes, it works. The service is currently hosted on Netlify: https://ogp-studio.netlify.app.

> Additionally, I think we can include more statistics in the rendered image like # of operations

Okay, I will add it.

apps/studio-next/src/helpers/parser.ts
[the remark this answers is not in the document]
KhudaDad414: They both serve the same purpose, so I guess you can look at both of them and combine them into a util function.

helios2003:
Added modifications:
- Using a package for detecting crawlers.
- Using ``sharp`` to optimize the image generation as mentioned here: https://nextjs.org/docs/messages/install-sharp.
- Add the dynamic behavior for ``url`` query param as well.

apps/studio-next/src/app/api/crawler/route.tsx
```diff
@@ -0,0 +1,100 @@
...
+       <html lang="en">
+       <head>
+        <meta charset="UTF-8">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>"AsyncAPI Studio"</title>
+        <meta property="og:title" content="AsyncAPI Studio" />
+        <meta property="og:description" content="Studio for AsyncAPI specification, where you can validate, view preview documentation, and generate templates from AsyncAPI document." />
+        <meta property="og:url" content="https://studio-next.netlify.app" />
+        <meta property="og:image" content="https://raw.githubusercontent.com/asyncapi/studio/master/apps/studio-next/public/img/meta-studio-og-image.jpeg" />
```
smoya, line 44: Can those values be grabbed from https://github.com/asyncapi/studio/blob/a428b15f6993687012396961e0032d14ee17ebee/apps/studio-next/src/app/page.tsx#L12 ?

helios2003: Done, @smoya

apps/studio-next/src/middleware.ts
```diff
@@ -0,0 +1,30 @@
...
+        return NextResponse.rewrite(new URL(`/api/crawler?url=${documentURL}`, request.url));
+      }
+    }
+  }
+  return res;
+}
+
+export const config = {
+  matcher: ['/:base64', '/:url'],
+}
```
KhudaDad414, line 30: Is this matcher configuration correct? I feel like it is matching to path parameters instead of query parameters that we use for url and base64.

helios2003: Oh yes, you are right, I didn't interpret the [documentation](https://nextjs.org/docs/app/building-your-application/routing/middleware#matching-paths) correctly, I will make the changes.

helios2003:
The PR is ready for review @KhudaDad414, @smoya.

helios2003:
Also, I'm unsure if the middleware code I wrote is the best way to achieve this functionality. Any suggestions or improvements would be appreciated.

apps/studio-next/src/app/api/crawler/route.tsx
```diff
@@ -0,0 +1,105 @@
...
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>${info.title}</title>
+        ${info.title ? `<meta property="og:title" content="${info.title}" />` : ''}
+        ${info.description ? `<meta property="og:description" content="${info.description}" />` : ''}
+        <meta property="og:image" content=${ogImageurl} />
+      </head>
+      </html>
+    `;
+    console.log(crawlerInfo);
```
smoya, line 94:
```suggestion
```

smoya:
@helios2003 Are you planning to move the OG Image generator service (Netlify Edge Function) into this PR?

helios2003:
Maintainers, can you please review the PR? I have addressed all the comments above and included the unit tests. Additionally, I had to change the version of ``@asyncapi/converter`` after this commit: https://github.com/asyncapi/converter-js/commit/b3592ef21a288231871080945b0b7c53b70adb32. 
I did not include these changes in the ``studio`` repository since we are soon switching to ``studio-next`` as the default version for ``studio`` based on this issue: https://github.com/asyncapi/studio/issues/1134

Amzani:
Some tests are still failing @helios2003

helios2003:
@Amzani the tests are failing in ``studio`` which I did not update as mentioned above.

KhudaDad414 (commented):
tests are faliling since you are installing a newer version of `jest-environment-jsdom` on `studio-next` app explicitly.

to resolve install the older version in `studio` app.
`pnpm install -D jest-environment-jsdom@27.5.1`

Shurtu-gal:
@smoya @helios2003 any status on this?

helios2003:
I can make changes to this PR, but I believe the sharing mechanism has changed a bit, so I need to adapt the code accordingly.

PR #518, opened 2022-11-30
refactor: switch to the nice-modal-react modal management

magicmatatjahu:
**Description**

- switch to the `@ebay/nice-modal-react` library that provides the fine modal management (better than base it on state).
- update `SettingsModal` with loading/updating the new `settings` state. 

That PR is a part of refactoring of the Studio to support multiple files.

PR #1135, opened 2024-09-16
feat: implement sharing of document through netlify blobs

### LINKED ISSUE #127: Share feature proposal
#### Reason/Context
The idea of this proposal is to add a "Share" button (or any component that fits) allowing the user to generate a URL that can be shared with anyone, and that will open the Studio but with the desired spec file loaded, representing that file in a certain moment in time, and loading any Studio setting the person who shared configured by then as well.

Sharing a URL that points to a reproducible Studio state is useful for several scenarios:

- Users can share AsycnAPI files without having to host them somewhere (like GH gists, etc), plus the fact you have all Studio amazing features out of the box.
- Tools that generate AsyncAPI files (like [Chatbot](https://github.com/asyncapi/chatbot), CLI...) could give a URL as a result instead of generating the code, or printing to stdout, etc.

You can see an example of such feature in Go (language) playground: https://play.golang.org/p/ENxYVMoY9VZ

#### Description
Let's see first a couple of mocks to illustrate the idea:
Mock share button

Mock modal

This is not as trivial as it might seem. I'm gonna add some context:

The [first idea](https://github.com/asyncapi/playground/issues/202) was to just accept all state (spec file, settings...) encoded as base64 in a query param. Then the studio will decode that value to obtain the state and draw it. 
The problem with this approach, as @magicmatatjahu [stated](https://github.com/asyncapi/playground/issues/202#issuecomment-947507773), is that each browser have their own limitations on the URL length (also with the length of each query parameter). Most of them, limiting the full URL to 2048 chars which is not a lot. E.g. [Streetlights MQTT example](https://github.com/asyncapi/spec/blob/master/examples/streetlights-mqtt.yml) has ~6000 chars already.

With such limitation in place, there are not a lot of solutions for solving this. 

#### Proposal
**TL;DR: Storing the state in an external storage. Downloading it later and import it into the Studio.** 

> [!IMPORTANT]  
> An alternative implementation, "cheapest" and simple-enough is mentioned in https://github.com/asyncapi/studio/issues/127#issuecomment-2074404179

Simple diagram

- Generating the URL will require passing the current state (AsyncAPI file + any settings) to an external service, which will generate a **UUID** (either based on the content like SHA256, or a UUID v(x). To be defined) and store the given state in a Storage (external database, cloud storage, etc). That **UUID** will be returned to the Studio and will be the only identifier needed to construct the shareable URL. Optionally, to use (or maybe create our own?) a URL shortener service for making the URL shorter.
- Loading a shared URL will require a call to an external service (same), which from a given **UUID**, will fetch the state from a Storage (external database, cloud storage, etc) and return it to the Studio, which will load such state.

Note: Other services like [Chatbot](https://github.com/asyncapi/chatbot/issues/26) could also use this service.

Obviously, this is only the big picture. Details should be clarify. Further investigation should be done, especially around the following topics among others:

- What storage should we use? How can we get budget for it?
- Shall we use https://github.com/asyncapi/server-api as server for making the sharing logic? Or rather a separate microservice that just owns such responsibility?
- Security concerns. Only available to our services.
- Monitoring seems now like a must, as it is a server giving a functionality to a public hosted tool. High Availability is a must as well.
- Which protocol to use? GRPC?
- Etc.

cc @magicmatatjahu @derberg

### LINKED ISSUE #1030: Long AsyncAPI document URL doesn't work
#### Describe the bug
When we have a long AsyncAPI doc and if we try to share it via the copied URL it gives HTTP 414 Error code which translates to URI being too long.
#### How to Reproduce
- Open AsyncAPI studio and choose Kraken Websocket example.
- After the example is generated click on the Share link which generates the URL of the document.
- Paste this in the browser.
- It gives an HTTP 414 error.
- Screen recording

https://github.com/asyncapi/studio/assets/100564488/50ac585f-8176-4a9e-a46c-086f52ad3434

#### Expected behavior
The document should be rendered as it is.

Shurtu-gal:
**Description**

- Added functionality for sharing of AsyncAPI documents through serverless functions and netlify blobs.
- The base64 version is still present in case someone don't wish the document to be stored at our netlify stores.

> [!IMPORTANT]
> I had tried implementing this through `edge functions` but turns out that support for npm modules is still sketchy out there leading to the following failure https://app.netlify.com/sites/share-studio/deploys/66e7f0045f2439ebfeadeae6#L150
>  More details: https://ntl.fyi/edge-functions-npm

**Related issue(s)**
Resolves #127
Resolves #1030


Shurtu-gal:
Could you please have a look at this @smoya @Amzani?

apps/studio-next/src/components/Editor/ShareButton.tsx
```diff
@@ -12,16 +12,13 @@ export const ShareButton: React.FunctionComponent<ShareButtonProps> = () => {
...
-          base64
-        )}`;
+        const url = await editorSvc.exportAsURL();
         await navigator.clipboard.writeText(url);
       }()),
       {
         loading: 'Copying URL to clipboard...',
         success: 'URL copied to clipboard!',
-        error: 'Failed to copy URL to clipboard.',
+        error: 'Failed to share document.',
```
Amzani, line 21:
```suggestion
        error: 'Failed to share the AsyncAPI document.',
```

apps/studio-next/src/components/Modals/ImportUUIDModal.tsx
```diff
@@ -0,0 +1,57 @@
+import { useState } from 'react';
+import toast from 'react-hot-toast';
+import { create } from '@ebay/nice-modal-react';
```
Amzani, line 3: I see that we've used this library without an ADR, It's fine.
Could we just add an ADR for it, to understand the reasoning behind it.

Amzani (commented):
Nice PR @Shurtu-gal 🚀 
Small things to fix in addition to conflict

apps/studio-next/src/components/Modals/ImportUUIDModal.tsx
Shurtu-gal: Sure, let me check in which commit it was introduced initially.

Shurtu-gal: @Amzani it was introduced in this PR https://github.com/asyncapi/studio/pull/518.

Amzani (approved):
LGTM

aeworxet:
@asyncapi/bounty_team

Issue #1138, opened 2024-09-17
Studio fell in production


aeworxet:
#### Describe the bug
Studio is not functional at https://studio.asyncapi.com with error in console.

#### How to Reproduce
Open https://studio.asyncapi.com.

![image](https://github.com/user-attachments/assets/ce04daa8-b629-4eb1-99de-b58e5a28bce6)

#### Expected behavior
Studio is expected to work.

aeworxet:
The bug is caused by code from https://github.com/asyncapi/converter-js introduced in

```
92cfd9a0877ca2f9483d8bca0b67d8f0479aca0e
```

A fast solution would be to revert and then figure out what happened.

@jonaslagoni @ItshMoh

aeworxet:
Take into account https://github.com/asyncapi/studio/pull/1094#issuecomment-2344700719, just in case.

aeworxet:
Another fast solution would be to temporarily hard-lock the `@asyncapi/converter` version in both `package.json`s:

```
./apps/studio/package.json
./apps/studio-next/package.json
```

```
"@asyncapi/converter": "1.5.0",
```

Choose which one is to your liking.

jonaslagoni:
Because the converter never had a requirement that the code must be able to run in the browser, I dont think the converter will get a rollback, unless majority wants it.

Studio needs to run the conversion logic in the backend and send it to the frontend, it's a problem because a dependency is no longer browser-compatible.

aeworxet:
Then, to get the Studio up and running right now, the only thing that is possible is to hard-lock the `@asyncapi/converter` on version `1.5.0` and after that start looking for a permanent solution.

Amzani:
We can go with this option @aeworxet

PR #1139, opened 2024-09-18
fix: temporarily hard-lock `@asyncapi/converter` on `v1.5.0` as a hotfix (#1138)

aeworxet:
This PR temporarily hard-locks the `@asyncapi/converter` on version `1.5.0`, to get the Studio up and running on production right now.
The permanent solution should be developed later.

Closes https://github.com/asyncapi/studio/issues/1138


Amzani (approved):
LGTM
Thanks @aeworxet for the fix

PR #1151, opened 2024-10-26
chore: add title and other metadata

Shurtu-gal:
**Description**

Currently title just shows the URL of the website which is not good.


Amzani (approved):
LGTM
As discussed in chat with @Shurtu-gal studio-next is failing as we still have Vercel config pointing to that dir, we can ignore it.
@fmvilas you could prolly remove that from Vercel
Thanks

PR #1150, opened 2024-10-23
ci: publish AsyncAPI Studio Docker image with additional version tag

fgreinacher:
**Description**

This PR extends the Docker publishing workflow so that an additonal version tag is published to Docker Hub. This already happened before the migration from `semantic-release` to `changesets`. I think it's desirable to bring it back because there are only super outdated versions on Docker Hub right now: https://hub.docker.com/r/asyncapi/studio/tags.

**Related issue(s)**

Let me know if I should create one.

:hammer_and_pick: with :heart: by [Siemens](https://opensource.siemens.com/)

PR #1160, opened 2025-01-08
fix(studio): fix Docker build

fgreinacher:
**Description**

The Docker publish workflow has been broken since the migration to Next.js and/or PNPM, see https://github.com/asyncapi/studio/actions/workflows/publish-docker-image.yml.

This MR adapts the Dockerfile to be compatible with the new setup.

**Related issue(s)**

Discovered while trying to validate https://github.com/asyncapi/studio/pull/1150

⚒️ with ❤️ by [Siemens](https://opensource.siemens.com/)

### DIFF
--- a/apps/studio/Dockerfile
+++ b/apps/studio/Dockerfile
@@ -9,7 +9,7 @@ RUN apk add --no-cache libc6-compat
 RUN apk update
 # Set working directory
 WORKDIR /app
-RUN npm install --global turbo
+RUN npm install --global turbo@1
 COPY . .
 RUN turbo prune --scope=@asyncapi/studio --docker
  
@@ -21,17 +21,17 @@ ARG BASE_URL_PLACEHOLDER
 RUN apk add --no-cache libc6-compat
 RUN apk update
 WORKDIR /app
- 
+RUN npm install --global pnpm@latest-10
+
 # First install the dependencies (as they change less often)
 
 COPY .gitignore .gitignore
 COPY --from=builder /app/out/json/ .
-COPY --from=builder /app/out/package-lock.json ./package-lock.json
-RUN PUPPETEER_SKIP_DOWNLOAD=true npm ci
- 
+RUN PUPPETEER_SKIP_DOWNLOAD=true pnpm install
+
 # Build the project
 COPY --from=builder /app/out/full/ .
-RUN PUBLIC_URL=${BASE_URL_PLACEHOLDER} npm run build:studio
+RUN PUBLIC_URL=${BASE_URL_PLACEHOLDER} NEXT_CONFIG_OUTPUT=export pnpm run build:studio
  
 
 FROM docker.io/library/nginx:1.25.5-alpine as runtime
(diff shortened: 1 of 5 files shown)


apps/studio/Dockerfile
```diff
@@ -9,10 +9,10 @@ RUN apk add --no-cache libc6-compat
 RUN apk update
 # Set working directory
 WORKDIR /app
-RUN npm install --global turbo
+RUN npm install --global turbo@1
```
fgreinacher, line 12: Need to pin the version because the config files here are not compatible with Turbo v2.

apps/studio/Dockerfile
```diff
@@ -21,32 +21,32 @@ ARG BASE_URL_PLACEHOLDER
 RUN apk add --no-cache libc6-compat
 RUN apk update
 WORKDIR /app
- 
+RUN npm install --global pnpm@latest-10
```
fgreinacher, line 24: As suggested in https://pnpm.io/installation#using-npm

apps/studio/Dockerfile
```diff
@@ -21,32 +21,32 @@ ARG BASE_URL_PLACEHOLDER
...
+
 # First install the dependencies (as they change less often)
 
 COPY .gitignore .gitignore
 COPY --from=builder /app/out/json/ .
-COPY --from=builder /app/out/package-lock.json ./package-lock.json
```
fgreinacher, line 29: This is not needed anymore.

apps/studio/tsconfig.json
```diff
@@ -28,7 +28,8 @@
         "./src/*",
         "./public/*"
       ]
-    }
+    },
+    "types": ["cypress"]
```
fgreinacher, line 32: Required so that the Next.js type checker is happy, see https://docs.cypress.io/app/tooling/typescript-support#Configure-tsconfigjson.

apps/studio/Dockerfile
```diff
@@ -21,17 +21,17 @@ ARG BASE_URL_PLACEHOLDER
...
-RUN PUPPETEER_SKIP_DOWNLOAD=true npm ci
- 
+RUN PUPPETEER_SKIP_DOWNLOAD=true pnpm install
+
 # Build the project
 COPY --from=builder /app/out/full/ .
-RUN PUBLIC_URL=${BASE_URL_PLACEHOLDER} npm run build:studio
+RUN PUBLIC_URL=${BASE_URL_PLACEHOLDER} NEXT_CONFIG_OUTPUT=export pnpm run build:studio
```
fgreinacher, line 34: Use [static HTML export](https://nextjs.org/docs/14/app/building-your-application/deploying#static-html-export) so that it works with the nginx-based setup.

This [has some limitations](https://nextjs.org/docs/14/app/building-your-application/deploying/static-exports#unsupported-features), but none of them seem to apply here at the moment.

A potential future improvement would be to re-engineer the docker setup, but I'd like to get it working again first.

Shurtu-gal:
Sorry for taking so long to come to this @fgreinacher. I will give this a look this weekend and test things out. Thanks for the help.

Shurtu-gal (approved):
LGTM 🚀

PR #1167, opened 2025-02-28
feat: additional changes for implementation of new preview command

### LINKED ISSUE #1166: Additional parameter to start studio using core package with `readOnly` parameter.
#### Reason/Context
   As discussed [here](https://github.com/asyncapi/cli/issues/1627#issuecomment-2682035401) onwards a new command `asyncapi start preview ` needs to be implemented for which studio has to be started in `readOnly` mode.

This should be done over existing implementation of mechanism of loading file and no extra websocket connections to be started and the `readOnly` should still only work with base64 string or url parameter or share when used outside of the to be implemented command `asyncapi start preview`.

#### Description
 An additional type in `IncomingMessage` needs to be added which when sent through existing websocket connection will allow to modify the state of frontend to `readOnly`.

   NO

neoandmatrix:
**Description**

This PR introduces the necessary changes for the implementation of  `readOnly` mode in studio when used with `asyncapi start preview` command.

Requirments:-

As the requirements of the implementation of this are following:-

- Starting of studio in it's minimal state when done through the to be implemented command `asyncapi start preview`.
- Other elements are disabled in this.
- Additon of websocket for syncing the  changes that come form file but not allowed to change anything from studio side.

These are also mentioned here https://github.com/asyncapi/cli/issues/1627#issuecomment-2730869866.

 Approach:-

To achieve the above requirements I tried a few different approaches as below :-

1. As current implementation of readOnly requires the file to be served in from either URL or base64 so one approach can be to start a mini file server and server the file from same. This approach is problematic as first starting the file server will consume more memory and second being that syncig through websocket will be problematic. Hence this was discarded.
2. One more approach can be to use current `start studio` command to set the readOnly state but this is also not feasible as first this will alter the current `start studio` command functionality and also disabling the editing will be challenging.

Now comes the approach which this PR presents. This PR introduces a new service for handling the `start preview` command. When studio is started through this first form the cli a confirmation message is sent before any message that this instance of studio was started with the `start preview` command and then only set the readOnly state in the studio. This verification ensures that the readOnly can only be set from this command or either from the other `url` approach. A check is also there to prevent this form setting the readOnly state again if it is already set through the presence of `url` and `readOnly` in the URL parameters which ensures that the normal behaviour is not affected. 

Also the pop up dialog that asks to change schema if the loaded schema is old one(like version 2.x or 1.x) as the state is readOnly so allowing changes is bit counter intuitive. 

**Related issue(s)**

Resolves #1166

### DIFF
--- a/.changeset/honest-candles-brush.md
+++ b/.changeset/honest-candles-brush.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/studio": patch
+---
+
+feat: additional changes for implementation of new preview command
--- a/apps/studio/src/components/Modals/ConvertToLatestModal.tsx
+++ b/apps/studio/src/components/Modals/ConvertToLatestModal.tsx
@@ -5,7 +5,7 @@ import { create } from '@ebay/nice-modal-react';
 import { ConfirmModal } from './ConfirmModal';
 
 import { useServices } from '../../services';
-import { useDocumentsState } from '../../state';
+import { useDocumentsState,appState } from '../../state';
 
 interface ConvertToLatestModal {
   convertOnlyToLatest: boolean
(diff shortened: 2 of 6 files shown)


neoandmatrix:
@Shurtu-gal Can you please have a look at the mentioned approach and changes.
If all good will proceed with this PR.

Thanks.

neoandmatrix:
@Shurtu-gal can you please have a look at the mentioned approach and changes.

If all good then will proceed with this.

Thanks.

neoandmatrix:
@Shurtu-gal any update on this?

Shurtu-gal:
Some other things also maybe need to be changed, like removing other views etc.
This can be a good starting point though. 

Have you tested it out?

neoandmatrix:
As mentioned in the approach this change will only affect the implementation of `asyncapi start preview` so to test this as per my understanding i might need to build this package locally and replace in cli codebase with the one installed via npm then test out the working of changes, right?.

Or one other way could be to test websockets via sending requests through postman or something.

@Shurtu-gal Let me know your thoughts on above mentioned approaches and would be testing based on that and if some other way is also possible please guide.

Thanks.

Shurtu-gal:
The first one is preferable. You can change package json to point to locally changed repo and start from there.

neoandmatrix:
As mentioned in the approach this change will only affect the implementation of `asyncapi start preview` so to test this as per my understanding i might need to build this package locally and replace in cli codebase with the one installed via npm then test out the working of changes, right?.

Or one other way could be to test websockets via sending requests through postman or something.

@Shurtu-gal Let me know your thoughts on above mentioned approaches and would be testing based on that and if some other way is also possible please guide.

Thanks.

Okay thanks.

Will be proceeding with same.

neoandmatrix:
> The first one is preferable. You can change package json to point to locally changed repo and start from there.

Okay, will proceed with same.

neoandmatrix:
@Shurtu-gal was trying to build the application but the builds are failing , and just to be sure that it's not due to changes i tried to build it form the official https://github.com/asyncapi/studio repository and even that is failing.

image

neoandmatrix:
@Shurtu-gal was trying to build the studio and linking it to cli for testing but after the build for studio is complete i am not able to understand which folder form `studio/build` to link to cli for testing this and also there seems to be some problem with build as it's only working once like for example if I do some changes and then build for first time build is successful and everything works , now but if i again just run the same build command `pnpm run build ` the build fails with nothing changed since last command. Also `pnpm run dev ` starts to fail also.

It would be very helpful to me if you could provide some guidance on these issues for me to test this implementation.

Thanks.

neoandmatrix:
> @Shurtu-gal was trying to build the studio and linking it to cli for testing but after the build for studio is complete i am not able to understand which folder form `studio/build` to link to cli for testing this and also there seems to be some problem with build as it's only working once like for example if I do some changes and then build for first time build is successful and everything works , now but if i again just run the same build command `pnpm run build ` the build fails with nothing changed since last command. Also `pnpm run dev ` starts to fail also.
> 
> It would be very helpful to me if you could provide some guidance on these issues for me to test this implementation.
> 
> Thanks.

@Shurtu-gal any help on this 😅 or suggestions on to what i tried.

neoandmatrix:
> You would need to give me some time. Running a bit busy right now 😕 

Oh sure sure, sorry if i caused any inconvenience. 😞

Shurtu-gal:
> The first one is preferable. You can change package json to point to locally changed repo and start from there.

@neoandmatrix you csn try this approach.

neoandmatrix:
> @Shurtu-gal was trying to build the studio and linking it to cli for testing but after the build for studio is complete i am not able to understand which folder form `studio/build` to link to cli for testing this and also there seems to be some problem with build as it's only working once like for example if I do some changes and then build for first time build is successful and everything works , now but if i again just run the same build command `pnpm run build ` the build fails with nothing changed since last command. Also `pnpm run dev ` starts to fail also.
> 
> It would be very helpful to me if you could provide some guidance on these issues for me to test this implementation.
> 
> Thanks.

Sure @Shurtu-gal i was also trying the first approach but as mentioned above build is not working correctly and hence the files generated are quite different from the currently used package.

Can you please name which folder/file should i link from build output.

Thanks.

Shurtu-gal:
This will go a little deeper. We have not updated cli, to latest studio hence problems with next has not come up. Need a deeper dive into this.

neoandmatrix:
> This will go a little deeper. We have not updated cli, to latest studio hence problems with next has not come up. Need a deeper dive into this.

Oh, then should i try once to upgrade it to latest in cli then maybe this can be implemented?

Also is issue with build actually there or something was wrong from my side?

neoandmatrix:
A demo of the testing of the above mentioned approach with the required changes in the cli locally. Please ignore logs in terminal those were for testing purpose only.


https://github.com/user-attachments/assets/4cc60929-5a04-4155-97ad-df40fe84c6f4

neoandmatrix:
Hey @Shurtu-gal sorry for a bit of inactivity on this, but have done the changes required for the `asyncapi start preview` command implementation and also update the approach along with a demo in above comments.

Approach : - https://github.com/asyncapi/studio/pull/1167#issue-2887032462
Demo :- https://github.com/asyncapi/studio/pull/1167#issuecomment-2773363637

Please review and let me know your thoughts on this. In my testing and verification everything is working perfectly fine. 
After this is approved will open the PR in the `cli` with the required changes to use this as per the above attached demo.

Thanks.

Shurtu-gal (dismissed):
LGTM! Very good work. Please be sure to bump the version in CLI after a release.

Shurtu-gal (approved):

Issue #1159, opened 2025-01-03
Studio Design System not working.


ShashwatPS:
#### Describe the bug
Package named asyncapi/studio-utils is not exporting its components and it is being used elsewhere in the project.

#### How to Reproduce
1. Clone the repo 
2. Run pnpm install
3. Run pnpm run dev or pnpm run ds
![Screenshot 2025-01-03 at 11 56 06 AM](https://github.com/user-attachments/assets/586c8141-7c52-4316-92af-5779a6265cce)

#### Expected behavior
After running the above commands, it shows up an error , could not resolve asynapi/studio-utils as its components are not exported in its package. 

If possible i would like to work on fixing this issue.

PR #1171, opened 2025-03-07
fix: import statement dropdown menu fixed

AkhileshJyotishi:
**Description**

- Fixed the import statement in `DropdownMenu.tsx`.
- Replaced the incorrect import from `@asyncapi/studio-utils` with the correct relative path `../../utils/src`.
- Ensured that the dropdown menu component loads correctly without dependency issues.

**Related issue(s)**  
Resolves: #1159


princerajpoot20:
@AkhileshJyotishi The issue #1159 is fixed with #1172
You can pick any other issue.

Issue #1104, opened 2024-05-28
Add the option of Generating docs in the main section 


AayushSaini101:
Reference: https://editor.swagger.io/ the Generator code section is present on the main screen, where as the in asyncapi studio devleoper need to click on three dots then need to click Generate doc/code.

For the easy accessiblity and increase the developer experience the generate code/docs option must be present in the main tab.

<img width="1892" alt="image" src="https://github.com/asyncapi/studio/assets/60972989/2b084cab-a2e1-42b1-b07c-7a0d3d5687d5">
<img width="1892" alt="image" src="https://github.com/asyncapi/studio/assets/60972989/67e1dc79-717c-4198-abdb-e1f5ee0e720e">

KhudaDad414:
Totally, I think we should create some icons for those options and then show the current options as a tooltip.

KhudaDad414:
@AayushSaini101 not sure if we can come up with an icon for each one, but it is easier to create four icons for each category and show a dropdown of those relevent options.
1) *import icon*.
   a) import from file
   b) import from url
   c) import from base64
2) *save icon*:
   a) save as yaml
   b) save as JSON
3) *convert icon*:
  a) convert to json/yaml
  b) convert to another AsyncAPI version
4) *generate code/doc icon*

abhavgoel:
@AayushSaini101 @KhudaDad414 should the the three dots be removed completely from the editor page?

KhudaDad414:
@abhavgoel yep. After we show the options in categories then nothing will remain for the the overflow menu.

ShashwatPS:
@KhudaDad414 You mean something like this ?

https://github.com/user-attachments/assets/bcfbbbc4-7314-493f-b8cb-fd382cf26fc8

If this is good enough, i can remove the borders which are there in the generate section and make a PR.

AayushSaini101:
> [@KhudaDad414](https://github.com/KhudaDad414) You mean something like this ?
> 
>  Screen.Recording.2025-01-04.at.11.29.25.AM.mov 
> If this is good enough, i can remove the borders which are there in the generate section and make a PR.

@ShashwatPS feel free to create a PR on this if you want thanks

AayushSaini101:
I will take this

aeworxet:
### Bounty Issue's service comment

Text labels: `bounty/2025-Q2`, `bounty/medium`, `bounty/coding`
[...]

PR #1184, opened 2025-03-31
feat: Added dropdowns in main menu bar over the Editor

abhavgoel:
**Description**

- Categorized and added options from the main dropdown into different dropdowns in the main bar over the editor
- Added icons which seemed most suitable according to the category

**Related issue(s)**
Related #1104

### DIFF
--- a/apps/studio/cypress.config.ts
+++ b/apps/studio/cypress.config.ts
@@ -2,7 +2,7 @@ import { defineConfig } from 'cypress';
 
 export default defineConfig({
   e2e: {
-    baseUrl: 'http://localhost:3001',
+    baseUrl: 'http://localhost:3000',
     retries: {
       runMode: 1,
       openMode: 1,
--- a/apps/studio/cypress/e2e/studio-ui.spec.cy.ts
+++ b/apps/studio/cypress/e2e/studio-ui.spec.cy.ts
@@ -112,25 +112,41 @@ describe('Studio UI spec', () => {
     cy.contains('Share link').should('be.visible');
   });
 
-  it('Button "Dropdown" should be visible in the UI', () => {
-    cy.get('[data-test="button-dropdown"]').should('be.visible');
-  });
-
-  it('Dropdown menu should contain 8 elements with predefined text', () => {
-    cy.get('[data-test="button-dropdown"]').click();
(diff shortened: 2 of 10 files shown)


Shurtu-gal (changes requested):
Please ensure the addition of the following things:
- There is no tooltip currently.
- Generation of code/docs is not working.

abhavgoel:
1. Where is tooltip needed?
2. The Generate code/docs is also not working in the prod.

Shurtu-gal:
> 1. Where is tooltip needed?

For the icons.

> 2. The Generate code/docs is also not working in the prod.

That's I mentioned addition of that in the issue 😄.
You would need to do a call to https://api.asyncapi.com (paths like `/v1/generate` and so on`. Would need to add testing for the same as well.

abhavgoel:
@Shurtu-gal current implementation makes a call to `localhost:3001`. Do I need to change the server url only? I tried but it still fails. I am not sure how the generator works

aeworxet:
@AayushSaini101

aeworxet:
@asyncapi/bounty_team

Shurtu-gal:
> Is this expected?

@abhavgoel no need to worry regarding generator of server-api not working for v3.0.0. Just go ahead and start working on adding the test suites. I will fix the api at the earliest.

AayushSaini101:
@abhavgoel if any case, I missed your message on the issues, kindly DM in the Slack. thanks, i will review your PR by EOD

AayushSaini101:
image
Looks Good, i guess we need to add test cases as well

abhavgoel:
Yes I am working on the test cases, I was adding it in the cypress module. What do you mean by `we need to add test cases as well apps/studio/src/services/server-api.service.ts for this file`? Do we need to test the api calls and responses too?

AayushSaini101:
> Yes I am working on the test cases, I was adding it in the cypress module. What do you mean by `we need to add test cases as well apps/studio/src/services/server-api.service.ts for this file`? Do we need to test the api calls and responses too?

if it is possible add test cases for the API call too

AayushSaini101 (approved):

Shurtu-gal (approved):
Thanks. LGTM 🚀.

Generation working with 2.6.0.

PR #1194, opened 2025-05-24
fix: build cache ignored for smaller bundle size

### LINKED ISSUE #1193: Reduce bundle and Package size of the studio.
   As form the discussion on the cli issue https://github.com/asyncapi/cli/issues/1752 and as per investigation in 
   same the bundling and package size of the studio is huge and thus increasing the overall size of other projects 
   using it as dependency.
- How will this change help?
  Reduce the overall package size to improve space and installation process.

#### Description
  Bundling and release pipelines needed to be updated to include only the required packages.
  No
- How could it be implemented/designed?
  Some reference :- https://nextjs.org/docs/app/guides/package-bundling

neoandmatrix:
**Description**

This pull request introduces the changes to update the `files` field in the `apps/studio/package.json` file to include only the necessary files and not the `cache` files in order to reduce the package size and bundle size.

Resolves #1193 

### Changes to `files` field in `package.json`:
* Replaced the single entry `/build` with a comprehensive list of specific build files and subdirectories, such as `/build/server`, `/build/standalone`, and `/build/static`, among others. This provides finer-grained control over the included build artifacts.

Screenshots of package size taken up in `cli` by `studio`  after removing cache files.

Screenshot From 2025-05-24 21-12-26
Screenshot From 2025-05-24 21-12-43
Screenshot From 2025-05-24 21-12-55
Screenshot From 2025-05-24 21-13-01
Screenshot From 2025-05-24 21-13-14

### DIFF
--- a/apps/studio/package.json
+++ b/apps/studio/package.json
@@ -15,13 +15,30 @@
   "author": {
     "name": "The AsyncAPI maintainers"
   },
-  "files": [
-    "/build",
-    "/public",
-    "./README.md",
-    "./LICENSE",
-    "./next.config.js"
-  ],
+ "files": [
+  "/build/server",
+  "/build/standalone",
+  "/build/static",
+  "/build/types",
+  "/build/app-build-manifest.json",
+  "/build/app-path-routes-manifest.json",
+  "/build/BUILD_ID",
+  "/build/build-manifest.json",
+  "/build/export-marker.json",
+  "/build/images-manifest.json",
+  "/build/next-minimal-server.js.nft.json",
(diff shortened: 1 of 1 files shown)


neoandmatrix:
Hey @Shurtu-gal done with changes. Sorry for delay. 
Please have a look . I have tested everything locally and works fine for me.

Thanks.

neoandmatrix:
> Tbh would still need to reduce cli node modules more. Let me mull on this a bit. You can do the same as well.

Sure, let's keep that issue open.
Will look further for optimization.

Thanks.

PR #1208, opened 2025-06-19
fix: support BASE_URL for studio docker builds

antoineauger:
**Description**

The support for the `BASE_URL` environment variable introduced within https://github.com/asyncapi/studio/pull/248 was sadly broken when the project migrated to Next.JS.

This PR is composed of two commits:
1. The first commit fixes the permissions when copying the static assets for the Docker image (in particular, the `logo-studio.svg` was not loading, even when running without a `BASE_URL` - see below)
2. The second commit adds the `basePath` Next.JS config option as documented within [the official doc](https://nextjs.org/docs/app/api-reference/config/next-config-js/basePath).

**Related issue(s)**

Related to https://github.com/asyncapi/studio/issues/249

**logo-studio.svg not loading due to nginx file permissions**

Screenshot from 2025-06-19 14-53-24

**PR test**

I built locally the image and deployed it on our k8s environment with `BASE_URL=/tools/asyncapi-studio` 

Unsaved Image 1

/cc @fgreinacher 

🛠️ with ❤️ by [Siemens](https://opensource.siemens.com/)

### DIFF
--- a/apps/studio/Dockerfile
+++ b/apps/studio/Dockerfile
@@ -43,7 +43,7 @@ ARG BASE_URL_PLACEHOLDER
 # https://github.com/nginxinc/docker-nginx/blob/master/entrypoint/docker-entrypoint.sh#L16.
 ARG ENTRYPOINT_SCRIPT=/docker-entrypoint.d/set-public-url.sh
 
-COPY --from=installer /app/apps/studio/build /usr/share/nginx/html/
+COPY --from=installer --chown=nginx:nginx /app/apps/studio/build /usr/share/nginx/html/
 # Add an entrypoint script that replaces all occurrences of the 
 # placeholder value by the configured base URL. If no base URL
 # is configured we assume the application is running at '/'.
--- a/apps/studio/Dockerfile
+++ b/apps/studio/Dockerfile
@@ -1,5 +1,5 @@
 # Use a UUID as placeholder value to have a unique string to replace. 
-ARG BASE_URL_PLACEHOLDER=189b303e-37a0-4f6f-8c0a-50333bc3c36e
+ARG BASE_URL_PLACEHOLDER=/189b303e-37a0-4f6f-8c0a-50333bc3c36e
 
 
 FROM node:18-alpine AS base
--- a/apps/studio/next.config.js
+++ b/apps/studio/next.config.js
@@ -36,6 +36,7 @@ const nextConfig = {
 
     return config;
   },
+  basePath: process.env.PUBLIC_URL ?? '',
   output: process.env.NEXT_CONFIG_OUTPUT ?? 'standalone',
   distDir: 'build'
 };
--- a/apps/studio/src/app/page.tsx
+++ b/apps/studio/src/app/page.tsx
@@ -34,8 +34,8 @@ export const metadata : Metadata = {
     images: ['/img/meta-studio-og-image.jpeg'],
   },
   icons: {
-    icon: '/favicon.ico',
-    apple: '/favicon-194x194.png',
+    icon: `${process.env.PUBLIC_URL ?? ''}/favicon.ico`,
+    apple: `${process.env.PUBLIC_URL ?? ''}/favicon-194x194.png`,
   },
 };
 


antoineauger:
I think this PR is ready for review, thanks 🏓

antoineauger:
@magicmatatjahu Do you think this PR makes sense? 

In the meantime, we built locally our own `asyncapi/studio` patch image but would really love to fix it upstream as well 😺

apps/studio/Dockerfile
```diff
@@ -31,22 +33,25 @@ RUN PUPPETEER_SKIP_DOWNLOAD=true pnpm install
...
 ARG ENTRYPOINT_SCRIPT=/docker-entrypoint.d/set-public-url.sh
 
-COPY --from=installer /app/apps/studio/build /usr/share/nginx/html/
+COPY --from=installer --chown=nginx:nginx /app/apps/studio/build /usr/share/nginx/html/
 # Add an entrypoint script that replaces all occurrences of the 
-# placeholder value by the configured base URL. If no base URL
+# placeholder values by the configured base URL. If no base URL
 # is configured we assume the application is running at '/'.
-RUN echo "find /usr/share/nginx/html/ -type f -print0 | xargs -0 sed -i \"s|${BASE_URL_PLACEHOLDER}|\${BASE_URL}|g\"" > $ENTRYPOINT_SCRIPT && chmod +x $ENTRYPOINT_SCRIPT
+RUN echo "find /usr/share/nginx/html/ -type f -print0 | xargs -0 sed -i \"s|${BASE_URL_PLACEHOLDER}|\${BASE_URL}|g\"" >> $ENTRYPOINT_SCRIPT && \
```
Shurtu-gal, line 53: Should the URL Placeholder be removed then?

Shurtu-gal (changes requested):

apps/studio/Dockerfile
antoineauger: @Shurtu-gal Thanks for the review. This was a good call, I simplified this PR and just pushed 72a94705ec306e687c04bc40280000f2590cbb09? 

Could you please have another look? (tested with success with and without a `BASE_URL` env variable)

apps/studio/Dockerfile
```diff
@@ -31,22 +33,23 @@ RUN PUPPETEER_SKIP_DOWNLOAD=true pnpm install
 
 # Build the project
 COPY --from=builder /app/out/full/ .
-RUN PUBLIC_URL=${BASE_URL_PLACEHOLDER} NEXT_CONFIG_OUTPUT=export pnpm run build:studio
+RUN NEXT_BASE_PATH=${BASE_PATH_PLACEHOLDER} NEXT_CONFIG_OUTPUT=export pnpm run build:studio
```
princerajpoot20, line 36: what is the reason behind changing the naming here? like, naming conventions is independent of `NEXT`

princerajpoot20 (changes requested):

apps/studio/Dockerfile
antoineauger: @princerajpoot20 Sorry about that, I initially got misled by my reading on [`NEXT_PUBLIC_` env variables](https://nextjs.org/docs/app/guides/environment-variables#bundling-environment-variables-for-the-browser) but I just noticed that we don't need any special-prefixed env variable here to set Next JS `basePath` 🙇🏻 

I rebased and pushed another version where:
- I kept the initial `BASE_URL_PLACEHOLDER` but modified it to start with `/` 
- I set Next JS `basePath` based on this value

This should also address https://github.com/asyncapi/studio/pull/1208#discussion_r2275476761 made by @Shurtu-gal 
Thanks for the reviews!

princerajpoot20 (changes requested):
Good work @antoineauger . Just left a minor suggestion.

antoineauger:
@princerajpoot20 I cleaned up the unnecessary changes, this should be ready now 👌🏻

princerajpoot20:
@Shurtu-gal PTAL

PR #1210, opened 2025-06-28
fix: Excluded templates that dont support asyncapi 3.0.0

### LINKED ISSUE #1192: Async API Studio fails to generate html / markdown doc
#### Describe the bug
The Async API studio is unable to generate HTML/Markdown doc

#### How to Reproduce
Try to generate HTML code using the sample template, error appears asking for latest version, even though the latest version is used

Image

#### Expected behavior
The Async API studio should generate HTML/Markdown doc

abhavgoel:
**Description**

- The generator does not support doc version 3.0.0 for these templates, as we get an error message when trying to generate the same
- Its better to disable these options from menu until we have support 
- This will reduce issues raised by users or any misconception.

Resolves #1192

### DIFF
--- a/apps/studio/src/components/Modals/Generator/GeneratorModal.tsx
+++ b/apps/studio/src/components/Modals/Generator/GeneratorModal.tsx
@@ -23,6 +23,9 @@ const unsupportedGenerators = [
   '@asyncapi/php-template',
   '@asyncapi/python-paho-template',
   '@asyncapi/ts-nats-template',
+  '@asyncapi/html-template',
+  '@asyncapi/java-template',
+  '@asyncapi/markdown-template',
 ];
 
 const renderOptions = (actualVersion: string) => {


Shurtu-gal:
Sorry, currently relocating so not able to devote much time

Regarding this, does none of the templates support v3?

abhavgoel:
As per code there were already some templates that dont support v3. The remaining ones that I added did not support v3. I confirmed with v2.6 and they were working fine

Shurtu-gal:
Did you test this on studio or using generator?

Because if you did this on studio, no v3 is working due to problems with server-api.

Another thing which can be done is to disable the whole v3 thing itself till server-api is not fixed.

Wdyt @Amzani

Shurtu-gal:
We can disable the generate option altogether for v3 till this is done: https://github.com/asyncapi/cli/issues/1790

Shurtu-gal:
Closing this in favor of #1220

Issue #1144, opened 2024-09-30
Generating code/docs results in error


haukepauke:
#### Describe the bug
When trying to generate a "HTML website" or "Markdown documentation" for the default project "Streetlights Kafka API 1.0.0" or any other project file an error message is shown: "Failed to generate."

This behaviour is similar for the version running on https:/studio.asyncapi.com and on a checkout of the master branch, running it locally (localhost:3000).

Browser console shows the error:

```POST http://localhost:3001/api/v1/generate net::ERR_CONNECTION_REFUSED``` (Chromium, Version 129.0.6668.70, Linux, 64 Bit)

```Quellübergreifende (Cross-Origin) Anfrage blockiert: Die Gleiche-Quelle-Regel verbietet das Lesen der externen Ressource auf http://localhost:3001/api/v1/generate. (Grund: CORS-Anfrage schlug fehl). Statuscode: ```
(Firefox, 130.0.1, Linux, 64-Bit).

#### How to Reproduce

1. Open studio.asyncapi.com
2. Click on the three dots in the upper right corner, choose "generate code/docs"
3. Pick f.e. "Markdown documentation", click "Generate"

![grafik](https://github.com/user-attachments/assets/8feeb12e-da43-4016-80ed-d31783f0c0ea)

#### Expected behavior
There should be no error message. Documentation should be generated.

catosaurusrex2003:
@aeworxet Please review my changes related to this [line](https://github.com/asyncapi/studio/commit/41365d42b8b0a7bee4197e966b5cc900875937b8#diff-7911c8d6c958d8d261602e2b153659dce570f122db8dc077ad1ed95912edd461R15) in the above PR

Amzani:
@aeworxet server-api is not working anyway with v3, so I suggest to directly support generator it in studio. 
cc @Shurtu-gal / @KhudaDad414

Shurtu-gal:
> @aeworxet server-api is not working anyway with v3, so I suggest to directly support generator it in studio. 
> cc @Shurtu-gal / @KhudaDad414 

Yeah this is the natural next step.

catosaurusrex2003:
still relevant

Shurtu-gal:
Hello folks,
This should be solved soon, I have migrated server-api to CLI. And generator v3 is also now supported.

PR #1220, opened 2025-09-18
feat: mitigate react2shell and add choice of generator version

### LINKED ISSUE #1192: Async API Studio fails to generate html / markdown doc

Shurtu-gal:
**Description**

- Currently CLI uses two versions: 1.17 and 2.8.3 for backwards compatibility. 
- So different templates use different generator versions.
- Upgrade next.js to 14.2.35.

**Related issue(s)**

Fixes #1192, #1144

### DIFF
--- a/.changeset/neat-toes-reflect.md
+++ b/.changeset/neat-toes-reflect.md
@@ -1,0 +1,5 @@
+---
+"@asyncapi/studio": minor
+---
+
+Mitigated React2Shell Vulnerability (Upgrade Next.js -> 14.2.35)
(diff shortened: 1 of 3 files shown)


abhavgoel:
I tested the flow and it works fine, now I can generate templates.

Shurtu-gal:
So html template requires you to use v2 generator which is indicated by the new button. Markdown is still not ported so requires older one.

Now regarding v3 docs, the problem was not with generator but with server-api as it was using older interfaces and code.

magicmatatjahu (approved):

PR #1242, opened 2025-12-22
chore(design-system): change ESLint invocation and ignore `node_modules`

batchu5:
This PR:

- Modifies the Design System's `lint` command to `eslint . --ext .ts --ext .tsx --ext .js --ext .jsx` to prevent parsing errors on Windows PowerShell (which treats commas as array separators, passing incorrect arguments to ESLint) and to ensure cross-shell compatibility.

- Explicitly adds the directory `node_modules` to the Design System's `.eslintignore` because the default ignore pattern is not honored.


princerajpoot20 (approved):
LGTM! @batchu5 Thanks for your contribution

PR #1331, opened 2026-07-09
fix: correct PUB/SUB label inversion in Navigation component

Gaurav77Kumar:
## Summary
Fixes the inverted PUB/SUB labels in the left panel Navigation component

Root Cause
In app/studio/src/components/Navigation.tsx;
- isReceive() block was showing "Pub" ❌
- isSend() block was showing "Sub" ❌

Fix
Swapped the labels to match correct AsyncAPI semantics:
- isReceive() -> Now shows "Sub" ✅
- isSend() -> now shows "Pub" ✅

### DIFF
--- a/apps/studio/src/components/Navigation.tsx
+++ b/apps/studio/src/components/Navigation.tsx
@@ -98,14 +98,14 @@ const OperationsNavigation: React.FunctionComponent<NavigationSectionProps> = ({
       if (operation.isReceive()) {
         channels.push(
           <li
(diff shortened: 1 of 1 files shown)


Shurtu-gal:
Will be worked under microgrant program. Thanks for the efforts.
I am available on slack, there are multiple more issues to be taken as well.

Gaurav77Kumar:
Thanks for the feedback, I will join Slack and look forward to contributing to more issues.

PR #1348, opened 2026-08-06
fix: issues related to current studio repo

Shurtu-gal:
**Description**

- Restored Studio build compatibility by rolling Next.js back to v14.
- Fixed layout flickering and pane-size jumps when toggling Navigation and Editor panels.
- Kept governance-setting labels on a single line.
- Display Pulsar operation channel names as `<persistence>://<tenant>/<namespace>/<address>` when complete, unambiguous bindings are available; otherwise retain the channel address.
- Extracted protocol-aware channel display formatting into an extensible helper registry.

**Related issue(s)**

Fixes #1334, #1332, #1326, #1278, #1255, #1330
Fixes #1333, #1335, #1292

PR #1360, opened 2026-08-18
fix: stop publishing unpublished studio-ui as a runtime dependency

princerajpoot20:
Fixes the CLI install failure caused by `@asyncapi/studio@1.4.0`.

That release declared a runtime dependency on `@asyncapi/studio-ui@0.5.0`, but `studio-ui` is a private workspace package and was never published to npm. This caused installs of `@asyncapi/cli` to fail with a 404.

This PR moves `@asyncapi/studio-ui` from `dependencies` to `devDependencies`, so Studio can still use it during build, but npm consumers will not try to install it.

Related: https://github.com/asyncapi/cli/issues/2273 
#1348


Shurtu-gal:
Yeah, sorry previous release auto added and bumped studio-ui.

Shurtu-gal (approved):

